PR #264, opened 2014-11-04
Accept '0' as a level which doesn't suffix the shortname.

mikewest:
Some specifications don't actually have levels. "Fetch", for instance. And "html5".

This patch accepts `"level": 0` as a hint that the shortname shouldn't be suffixed. An alternative approach would be to make the `level` attribute optional, but since you'd explicitly made it required, I'm not sure you'd accept that patch. :)

WDYT?

### DIFF
--- a/bikeshed/__init__.py
+++ b/bikeshed/__init__.py
@@ -555,7 +555,10 @@ def transformAnchors(lines, doc, **kwargs):
             if not checkTypes(anchor, key, field, basestring, int):
                 continue
             anchor[field] = unicode(anchor[field]).strip() + "\n"
-        anchor['spec'] = "{0}-{1}\n".format(anchor['shortname'].strip(), anchor['level'].strip())
+        if anchor['level'].strip() != '0':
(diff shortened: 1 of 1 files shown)


mikewest:
I dropped this on the floor long enough that I have no idea what the complaint was. If it crops up again, I'll file a new PR. :)

tobie:
I'm running the command line tool with the `--force` option to allow building without a level. I'm not sure why levels are required at level 1.

tabatkins:
Because it's still leveled (level 1) and if you ever publish a second level, you'll want to reserve the unleveled name for referring to the "most recent" version.  We've found this to be good practice in the CSSWG, and Bikeshed generally tries to encode good practices.

tobie:
Well, can't you still do that when you start working on level 2?

tabatkins:
If you're willing to go back and repub the old one, sure.  That's effort that most people won't do, and they shouldn't have to remember to do so.

tobie:
That makes sense.

Issue #319, opened 2015-01-08
"multiple-word" is not equal to "multiple word"


mikewest:
Bikeshed generates the same ID for both `<dfn>source expression</dfn>` and `<dfn>source-expression</dfn>`; it's arguable that I shouldn't have defined both those terms, but I did, and it broke. :(

Perhaps the latter could canonicalize to `source--expression`, since it's probably less common than the former?

mikewest:
(ID generation aside, though, you're doing the right thing with linking, which is great!)

tabatkins:
I added an argument to simplifyText to cause this dashes-become-underscores behavior, and now invoke it for "dfn" type definitions, and headings.

fantasai:
This change broke links on my specs. Also underscores suck for URLs. I don't think we should make things worse for everybody just because somebody defined "foo-bar" and "foo bar" as different terms and wants that to automatically work out.

tabatkins:
Why do underscores "suck"?  Why is anyone caring about the look of the ID in the first place?  That's what auto-linking is for, and making it work automatically for more things is more better.

fantasai:
Because they blend in with underlines, which is commonly used to format links. (Also they're harder to type.) Auto-linking is great for things that auto-link, but people cut and paste URLs to specs all over the place.

Also, your change probably broke a lot of links into our specs from outside our auto-generated ecosystem. Changing it break would break some newer links, but would fix those older ones (of which I'm sure there are more).

fantasai:
Another reason not to make this change is that a term might change from hyphenated to unhyphenated or vice versa, and that very superficial change would also break links.

fantasai:
For the case where someone has created "foo-bar" and "foo bar" as distinct terms, that's very unusual and can be handled by manually specifying the ID. I think the other cases here, for which reverting this change would be beneficial, are more common.

mikewest:
If I'm the only one using this, then back it out and I'll deal with it.

That said, I use this all over the place. The `Content-Security-Policy` HTTP response header is not the same as the "Content Security Policy" concept, the `source-expression` ABNF rule name is not the same as the "source expression" concept, and so on. It surprises me that it hasn't bitten anyone else.

tabatkins:
All right, I'm gonna go ahead and revert this.  Sorry about that, Mike.  I'll give you time to set your IDs manually; just ping this thread when you're ready.

mikewest:
[...]

PR #320, opened 2015-01-08
Convert '-' to '_' when generating IDs.

mikewest:
Closes tabatkins/bikeshed#319.

### DIFF
--- a/bikeshed/config.py
+++ b/bikeshed/config.py
@@ -211,6 +211,7 @@ def reSubObject(pattern, string, repl=None):
 def simplifyText(text):
     # Remove anything that's not a name character.
     text = text.strip().lower()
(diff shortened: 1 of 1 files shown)


mikewest:
WDYT, @tabatkins?

tabatkins:
Hmm, this'll change IDs for a bunch of existing CSS properties/keywords/functions, to an arguably worse form.

I can see this change for "dfn" type definitions, though.

mikewest:
Oh. Hrm. I was only looking at my specs, and kinda didn't think about CSS. Suggestions?

tabatkins:
Fixed this in a different way, per #319. Closing this PR.

mikewest:
Works for me, thanks!

PR #326, opened 2015-01-17
Add `((.../...))` autolink shortcuts for elements/attributes.

### LINKED ISSUE #323: Autolink shorthand for elements and element attributes.
I find myself writing `<code><a element-attr for="...">...</a></code>` more often than I'd like. It would be nice if there was a shorthand autolink for this kind of thing. If you hadn't used up ALL THE SPECIAL CHARACTERS already for boring old CSS, I'd have a reasonable suggestion for syntax. As-is, I will simply lament that `<<...>>` has already been swiped. That said, nothing matches `<iframe@src>` yet.  Or `{{iframe@src}}`, I guess. Or anything else?

mikewest:
This patch adds the fairly lame double-parens shortcut for elements and
element attributes. That is `((element))` will produce a link to the
element named 'element', and `((element/attribute))` will produce a link
to the attribute 'attribute' for the element 'element'.

Closes #323.

### DIFF
--- a/bikeshed/__init__.py
+++ b/bikeshed/__init__.py
@@ -1459,6 +1459,12 @@ class CSSSpec(object):
             return E.code({"class":"idl"},
                 E.a({"data-link-type":linkType, "for": match.group(1)}, match.group(2)))
 
+        elementRe = re.compile(r"<{(?:(\w+)/)?(\w+)}>")
+        def elementReplacer(match):
+            linkType = "element" if match.group(1) is None else "element-attr"
+            return E.code({},
+                E.a({"data-link-type":linkType, "for": match.group(1)}, match.group(2)))
+
         def transformElement(parentEl):
             processContents = isElement(parentEl) and not isOpaqueElement(parentEl)
             if not processContents:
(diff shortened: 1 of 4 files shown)


mikewest:
I stripped out some of the complexity around the `!!` syntax, as I don't think we need it for attributes and elements. WDYT, @tabatkins?

mikewest:
Ping. Crazy? Not crazy?

Would `<{.../...}>` be better or worse?

tabatkins:
Sorry for delay - I don't look at email over the weekend.

The code looks fine to me at a glance.  I'm fine with whatever shorthand you like more, just decide on one and I'll merge.

And yes, no need for the `!!` stuff - this shorthand is fully typed already from the syntax, so there's never a need to specify the type explicitly.

domenic:
Honestly I kind of like `<{.../...}>` if that counts for anything. I am pretty unlikely to be using this syntax though so don't take me too seriously :).

tabatkins:
@mikewest Do you prefer the <{foo}> syntax as well?  If so, I'll pull these edits in and switch it over.

mikewest:
Updated the patch with the `<{.../...}>` syntax. WDYT?

PR #338, opened 2015-01-29
<{.../...}> syntax ought to support all valid attribute/element characters.

mikewest:
Attribute can contain '-', for example. Who knew?

### DIFF
--- a/bikeshed/__init__.py
+++ b/bikeshed/__init__.py
@@ -1502,7 +1502,7 @@ class CSSSpec(object):
             return E.code({"class":"idl"},
                 E.a({"data-link-type":linkType, "for": match.group(1)}, match.group(2)))
 
-        elementRe = re.compile(r"<{(?:([^ \"'>/=]+)/)?([^ \"'>/=]+)}>")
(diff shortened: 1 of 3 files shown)


bikeshed/__init__.py
```diff
@@ -1502,7 +1502,7 @@ def idlReplacer(match):
             return E.code({"class":"idl"},
                 E.a({"data-link-type":linkType, "for": match.group(1)}, match.group(2)))
 
-        elementRe = re.compile(r"<{(?:(\w+)/)?(\w+)}>")
+        elementRe = re.compile(r"<{(?:([^ \"'>/=]+)/)?([^ \"'>/=]+)}>")
```
tabatkins: This regex is capable of matching across two element defs on the same line, because it's too liberal.  Just do `[\w-]+`; it's a pattern I use a bunch all throughout.

mikewest: Can it? It would stop after hitting a `/` or `>`, right?

But whatever. :) `\w-` is probably clearer.

mikewest:
Fixed?

Issue #361, opened 2015-02-13
Feature request: Autogenerate a list of external references.


mikewest:
Web Bluetooth does a nice job presenting the terminology brought in from other specs: https://webbluetoothcg.github.io/web-bluetooth/#terminology (yay, @jyasskin!). I've done similar things in various WebAppSec specs (https://w3c.github.io/webappsec/specs/upgrade/#terms-defined-by-reference, for instance). Currently, these lists are lovingly hand-crafted, which is somewhat absurd. :)

@tabatkins: How would you feel about a patch to bikehsed that captured the references used to generate a document (probably by making `ReferenceManager::getRef` return a reference rather than a URL, and storing a de-duped list on the `CSSSpec` object (which we should totally rename to `WebAppSecSpec` or something :) )?

jyasskin:
Heh, my list is a workaround for https://github.com/w3c/respec/issues/332 not being finished.

tabatkins:
Yeah, this seems quite reasonable. Only problem is that there's no real mapping between anchor spec names and biblio entries; they should quite often coincide, but won't always.  Though I suppose we can generate at least a minimal biblio entry from the specs database when there's no match.  You could also supply a custom biblio entry if desired.

And yeah, `CSSSpec` is one of the few remaining things pointing out that I clearly designed this for the CSSWG at first. ^_^

tabatkins:
Closing this, since the PR was merged.

Also: just landed the ability to auto-generate biblio entries from spec data, if the shortname isn't in the biblio data.  The spec data should match the autolink data, except for manually-added anchor data, so you should get all the biblio entries filling in now.

PR #366, opened 2015-02-16
Add external references to the Index.

mikewest:
This patch splits a document's index into two sections, the first
listing terms defined directly in the document, the second listing terms
defined elsewhere. The latter are grouped by specification, in the hopes
of making references clear for readers.

To do so, this patch changes `ReferenceManager::getRef` to return a
reference rather than a URL. When we process autolinks, we grab the
relevant references, and store them in a new `externalRefsUsed` property
on the `CSSSpec` object itself. In `addIndexSection`, we then walk
through the external references in order to build up a list of
definitions.

This is a first pass at tabatkins/bikeshed#361. There will be followups.

### DIFF
--- a/bikeshed/ReferenceManager.py
+++ b/bikeshed/ReferenceManager.py
@@ -203,7 +203,7 @@ class ReferenceManager(object):
         return self.queryRefs(text=text, linkType=linkType, status="local", linkFor=linkFor)[0]
 
     def getRef(self, linkType, text, spec=None, status=None, linkFor=None, error=True, el=None):
-        # If error is False, this function just shuts up and returns a url or None
+        # If error is False, this function just shuts up and returns a reference or None
         # Otherwise, it pops out debug messages for the user.
 
         # 'maybe' links might not link up, so it's fine for them to have no references.
@@ -257,15 +257,15 @@ class ReferenceManager(object):
             # Allow foo(bar) to be linked to with just foo(), but only if it's completely unambiguous.
             textPrefix = text[:-1]
             candidates, _ = self.queryRefs(linkType=linkType, status="local", linkFor=linkFor)
-            candidateRefs = list(set(c for c in candidates if c.text.startswith(textPrefix)))
+            candidateRefs = {c.url: c for c in candidates if c.text.startswith(textPrefix)}.values()
             if len(candidateRefs) == 1:
                 return candidateRefs[0]
             # And repeat for non-locals
             candidates, _ = self.queryRefs(linkType=linkType, spec=spec, status=status, linkFor=linkFor, export=export, ignoreObsoletes=True)
-            candidateRefs = list(set(c for c in candidates if c.text.startswith(textPrefix)))
+            candidateRefs = {c.url: c for c in candidates if c.text.startswith(textPrefix)}.values()
             if len(candidateRefs) == 1:
                 return candidateRefs[0]
-            if len(urls) > 1 and zeroRefsError:
+            if len(candidateRefs) > 1 and zeroRefsError:
                 die("Too many possible '{0}' targets to disambiguate. Please specify the names of the required args, like 'foo(bar, baz)'.", text)
 
         if failure and linkType in ("argument", "idl") and linkFor is not None and linkFor.endswith("()"):
@@ -273,12 +273,12 @@ class ReferenceManager(object):
             candidateFor, _, forPrefix = linkFor.partition("/") if "/" in linkFor else (None, None, '')
             forPrefix = forPrefix[:-1]
             candidates, _ = self.queryRefs(linkType="functionish", status="local", linkFor=candidateFor)
-            localRefs = list(set(c for c in candidates if c.text.startswith(forPrefix)))
+            localRefs = {c.url: c for c in candidates if c.text.startswith(textPrefix)}.values()
             if len(localRefs) == 1:
                 return localRefs[0]
             # And repeat for non-locals
             candidates, _ = self.queryRefs(linkType="functionish", spec=spec, status=status, linkFor=candidateFor, export=export, ignoreObsoletes=True)
-            remoteUrls = list(set(c for c in candidates if c.text.startwith(forPrefix)))
+            remoteUrls = {c.url: c for c in candidates if c.text.startswith(textPrefix)}.values()
             if len(remoteRefs) == 1:
                 return remoteRefs[0]
             if zeroRefsError and (len(localRefs) or len(remoteRefs)):
@@ -618,8 +618,6 @@ class RefWrapper(object):
     def __getattr__(self, name):
         if name == "for_":
             name = "for"
-        if name not in self.ref:
-            return None
         val = self.ref[name]
         if isinstance(val, basestring):
             val = val.strip()
--- a/bikeshed/ReferenceManager.py
+++ b/bikeshed/ReferenceManager.py
@@ -223,7 +223,7 @@ class ReferenceManager(object):
         # Local refs always get precedence, no matter what.
         localRefs = self.getLocalRef(linkType, text, linkFor, el)
         if len(localRefs) == 1:
-            return localRefs[0].url
+            return localRefs[0]
         elif len(localRefs) > 1:
             if error:
                 warn("Multiple possible '{0}' local refs for '{1}'.\nArbitrarily chose the one with type '{2}' and for '{3}'.",
@@ -233,7 +233,7 @@ class ReferenceManager(object):
                      "' or '".join(localRefs[0].for_))
                 if text == "encoding":
                     warn("{0}", config.printjson(localRefs))
-            return localRefs[0].url
+            return localRefs[0]
 
         # Take defaults into account
         if (not spec or not status):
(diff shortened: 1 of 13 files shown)


mikewest:
WDYT of this approach, Tab?

bikeshed/ReferenceManager.py
```diff
@@ -257,14 +257,14 @@ def getRef(self, linkType, text, spec=None, status=None, linkFor=None, error=Tru
             # Allow foo(bar) to be linked to with just foo(), but only if it's completely unambiguous.
             textPrefix = text[:-1]
             candidates, _ = self.queryRefs(linkType=linkType, status="local", linkFor=linkFor)
-            urls = list(set(c.url for c in candidates if c.text.startswith(textPrefix)))
-            if len(urls) == 1:
-                return urls[0]
+            candidateRefs = list(set(c for c in candidates if c.text.startswith(textPrefix)))
```
tabatkins: This needs to preserve the url-uniquifying behavior.  There are several instances where different refs have the same url, so warning about multiple refs is just an annoyance.  This can be done by switching to a url-keyed dict that only preserves one of the references, then calling `list(refs.values())`.

bikeshed/ReferenceManager.py
```diff
@@ -257,14 +257,14 @@ def getRef(self, linkType, text, spec=None, status=None, linkFor=None, error=Tru
...
+            if len(candidateRefs) == 1:
+                return candidateRefs[0]
             # And repeat for non-locals
             candidates, _ = self.queryRefs(linkType=linkType, spec=spec, status=status, linkFor=linkFor, export=export, ignoreObsoletes=True)
-            urls = list(set(c.url for c in candidates if c.text.startswith(textPrefix)))
-            if len(urls) == 1:
-                return urls[0]
+            candidateRefs = list(set(c for c in candidates if c.text.startswith(textPrefix)))
```
tabatkins: And same here, obvs.

bikeshed/ReferenceManager.py
```diff
@@ -273,21 +273,28 @@ def getRef(self, linkType, text, spec=None, status=None, linkFor=None, error=Tru
             candidateFor, _, forPrefix = linkFor.partition("/") if "/" in linkFor else (None, None, '')
             forPrefix = forPrefix[:-1]
             candidates, _ = self.queryRefs(linkType="functionish", status="local", linkFor=candidateFor)
-            localUrls = list(set(c.url for c in candidates if c.text.startswith(forPrefix)))
-            if len(localUrls) == 1:
-                return localUrls[0]
+            localRefs = list(set(c for c in candidates if c.text.startswith(forPrefix)))
```
tabatkins: And here.

bikeshed/ReferenceManager.py
```diff
@@ -273,21 +273,28 @@ def getRef(self, linkType, text, spec=None, status=None, linkFor=None, error=Tru
...
+            if len(localRefs) == 1:
+                return localRefs[0]
             # And repeat for non-locals
             candidates, _ = self.queryRefs(linkType="functionish", spec=spec, status=status, linkFor=candidateFor, export=export, ignoreObsoletes=True)
-            remoteUrls = list(set(c.url for c in candidates if c.text.startwith(forPrefix)))
-            if len(remoteUrls) == 1:
-                return remoteUrls[0]
-            if zeroRefsError and (len(localUrls) or len(remoteUrls)):
+            remoteUrls = list(set(c for c in candidates if c.text.startwith(forPrefix)))
```
tabatkins: And here. ^_^

bikeshed/ReferenceManager.py
```diff
@@ -273,21 +273,28 @@ def getRef(self, linkType, text, spec=None, status=None, linkFor=None, error=Tru
...
+            if len(remoteRefs) == 1:
+                return remoteRefs[0]
+            if zeroRefsError and (len(localRefs) or len(remoteRefs)):
                 die("Too many possible method targets to disambiguate '{0}/{1}'. Please specify the names of the required args, like 'foo(bar, baz)', in the 'for' attribute.", linkFor, text)
 
         if failure == "text" or failure == "type":
             if spec and spec in self.anchorMacros:
                 # If there's a macro registered for this spec, use it to generate a ref.
-                return anchorMacros[spec].url + config.simplifyText(text)
+                return {
```
tabatkins: Actually I think this can be removed?  I don't believe I support anchor macros right now.  It was a planned feature earlier, when `<pre class=anchors>` had a much more annoying and verbose JSON structure.

bikeshed/ReferenceManager.py
```diff
@@ -604,6 +618,8 @@ def __init__(self, text, ref):
     def __getattr__(self, name):
         if name == "for_":
             name = "for"
+        if name not in self.ref:
```
tabatkins: Why add this?  It breaks the standard "raise an error if the name doesn't exist" behavior.

bikeshed/__init__.py
```diff
@@ -746,14 +746,28 @@ def processAutolinks(doc):
...
                               error=(linkText.lower() not in doc.md.ignoredTerms))
-        if url is not None:
-            el.set('href', url)
+        # Capture the reference (and ensure we add a biblio entry) if it
+        # points to an external specification. We check the spec name here
+        # rather than checking `status == "local"`, as "local" refs include
+        # those defined in `<pre class="anchor">` datablocks, which we do
+        # want to capture here.
+        if ref and ref.spec is not None and ref.spec is not "" and ref.spec != doc.refs.specVName:
+            if ref.spec not in doc.externalRefsUsed:
```
tabatkins: Use defaultdict().  These kinds of checks are verbose, error-prone, and avoidable.

mikewest: `ref` isn't a dictionary. It's a WrappedRef.

bikeshed/ReferenceManager.py
mikewest: Not sure. I'm sure I had a good reason at the time, though! Dropping...

bikeshed/ReferenceManager.py
mikewest: `refs.values()` alone would be enough, right?

mikewest:
How does this look?

mikewest:
It successfully builds WebAppSec specs, with the exception of a mismatch between references' `css-images-3` and the specref name of `css3-images`. Same for `css-images-4` and `css4-images`. There doesn't seem to be a clean way to canonicalize that. :(

tabatkins:
There isn't, unfortunately.  Those are bugs in SpecRef which I think will eventually canonicalize themselves as we publish new drafts with the proper shortnames.

tabatkins:
I'll review the rest of the updated diff in a few hours, but I note that we have *one* failing test, due to a difference in editor name?  What's up with that?

mikewest:
Hrm. Apparently Travis is picking W3C HTML5, and locally I'm getting WHATWG (or vice versa, perhaps). :(

bikeshed/__init__.py
```diff
@@ -1248,6 +1262,7 @@ def preprocess(self):
         processHeadings(self, "all") # again
         addTOCSection(self)
         addSelfLinks(self)
+        processBiblioLinks(self) # again, as biblio links could be added by addIndexSection().
```
tabatkins: Actually, can it just be moved here?  Running once is better than running twice.

tabatkins:
All right, only one line comment, and the failing test, and I'll be ready to merge.  Thanks!

tabatkins:
Re: test, that's... strange.

tabatkins:
Ignore the failure, then. I'll debug that separately.

bikeshed/ReferenceManager.py
mikewest: Missed this on the previous pass. If it's not in use, I'm happy to kill it in a followup patch.

bikeshed/__init__.py
mikewest: We need to run it before we run `addReferencesSection`. I'll see if I can find an ordering that works. If you don't mind reversing the references section and the index, we could do `addIndex -> processBiblio -> addReferences`?

tabatkins: Yeah, the ordering there is unimportant. Go for it.

(I still intend to some day add an explicit ordering mechanism, so the important orderings become more visible, but meh.)

mikewest: Hrm. Processing `biblio` links needs to happen before `processAutoLinks`, as they're then transformed such that the latter doesn't attempt to look them up.

mikewest: (Fiddling)

mikewest: No, not this change. We'd need to rework `processAutoLinks` a bit.

If we move `processBiblioLinks` down after `processAutoLinks`, then the latter grabs all the biblio links and attempts to resolve them against the anchors list. Since there's no anchor named HTML5, it throws up it's hands in disgust and emits a fatal error.

We could exclude biblio links from the `findAll` call at the top of `processAutoLinks`, I suppose. I'll try that to see how it looks.

tabatkins: Yeah, excluding biblio links in `processAutoLinks` works for me.

mikewest: That's not too bad, actually. I'll throw another patch at you shortly; unfortunately, it involves rebasing a lot of tests. :/

mikewest:
Ok. Travis is happy(ish), and I've poked at `processAutoLinks` to exclude biblio-type links, which lets us call `processBiblioLinks` just once. WDYT?

tabatkins:
Yeah, let's do it.  Thanks!

Issue #254, opened 2014-10-29
Linking to W3C's HTML spec.


mikewest:
At some point between September and now, the autolinking mechanism flipped from pointing at W3C's HTML spec to the WHATWG's HTML spec. The entire discussion of quality and up-to-dateness to the side, it's unlikely that I can get anything through WebAppSec using all WHATWG specs all the time.

I suspect that most people using Bikeshed actually _want_ to link to WHATWG's HTML, but there might be room for a pref? Or for something like `Link Defaults: HTML` vs `Link Defaults: HTML5` (I'm a bit surprised that the latter links to WHATWG, honestly; that's not it's name)?

If I'm the only one who needs W3C references, I'd be perfectly happy generating the `anchors.data` file myself if there's some documentation about how to go about doing it. :)

tabatkins:
Closing because you've presumably been publishing with WHATWG HTML references for a while; CSS certainly has.  And W3C's HTML has become even more of a garbage fire than it was before.

PR #368, opened 2015-02-25
Add external issue tracker links to issues.

mikewest:
This patch adds a "ISSUE(...): " syntax to auto-generated, inlined
issues which simplifies the process of linking out to topics that are
being discussed on a remote issue tracker.

I considered a GitHub-specific syntax, but putting a template URL in the
document's metadata seemed simpler and more inclusive.

### DIFF
--- a/bikeshed/__init__.py
+++ b/bikeshed/__init__.py
@@ -803,8 +803,13 @@ def processIssues(doc):
         remoteIssueID = el.get('data-remote-issue-id')
         if remoteIssueID:
             del el.attrib['data-remote-issue-id']
-            if doc.md.issueTrackerTemplate:
+            githubMatch = re.match(r"([^/]+)/([^#]+)#(\d+)", remoteIssueID)
+            remoteIssueURL = None
+            if githubMatch:
+                remoteIssueURL = "https://github.com/{0}/{1}/issues/{2}".format(githubMatch.group(1), githubMatch.group(2), githubMatch.group(3))
+            elif doc.md.issueTrackerTemplate:
                 remoteIssueURL = doc.md.issueTrackerTemplate.format(remoteIssueID)
+            if remoteIssueURL:
                 appendChild(el, " ", E.a({"href": remoteIssueURL }, "<" + remoteIssueURL + ">"))
     dedupIds(doc, findAll(".issue", doc))
 
--- a/bikeshed/__init__.py
+++ b/bikeshed/__init__.py
@@ -803,7 +803,7 @@ def processIssues(doc):
         remoteIssueID = el.get('data-remote-issue-id')
         if remoteIssueID:
             del el.attrib['data-remote-issue-id']
-            githubMatch = re.match(r"([^/]+)/([^#]+)#(\d+)", remoteIssueID)
+            githubMatch = re.match(r"\s*([\w\-_]+)/([^\w\-_]+)#(\d+)\s*$", remoteIssueID)
             remoteIssueURL = None
             if githubMatch:
                 remoteIssueURL = "https://github.com/{0}/{1}/issues/{2}".format(githubMatch.group(1), githubMatch.group(2), githubMatch.group(3))
--- a/bikeshed/__init__.py
+++ b/bikeshed/__init__.py
@@ -803,7 +803,7 @@ def processIssues(doc):
         remoteIssueID = el.get('data-remote-issue-id')
         if remoteIssueID:
             del el.attrib['data-remote-issue-id']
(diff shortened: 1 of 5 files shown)


mikewest:
What do you think about this, @tabatkins? In short: some, but not all, of my specs' inline issues are being discussed on GitHub. I'd like to link to them without thinking too hard about it.

I like the issue syntax in this patch (`ISSUE(123):`), but I'm not at all convinced that "Issue Tracker Template" makes sense, nor am I really thrilled with the link syntax (which is `<http://issue-tracker.test/1234>` at the end of the issue).

Is this a crazy idea? If not, suggestions for improvements welcome!

tabatkins:
Overall, I agree with this! It sounds like a good addition, and it's a reasonable syntax.  Some nits:

1. In the rest of the codebase I explicitly use `.format()`, not `%`, to format strings, and that's also recommended Pythonic practice.  Using that here also means you have the imo clearer `http://example.com/issue{0}` syntax; people familiar with brace-based templating languages will be familiar with this.

2. Do we want multiple arguments?  I can see a spec referencing an issue in a *different* github repo, for example.  (In CSS we sometimes have issues scattered between Tracker, Bugzilla, and GitHub.) This might be handleable just by allowing multiple named templates, so you can refer to one by name and still just pass the issue number.  We can probably defer this.

3. The display syntax seems fine to me; we can tweak later if we come up with something better.

mikewest:
1. Sounds fine to me.

2. Since we're focusing on specs, it might be reasonable to just have syntax for those three options. I don't think I'd need that for any specs I'm working on, but I can see how it would be valuable. What probably _would_ be valuable is linking to other GitHub repositories (e.g. `ISSUE(tabatkins/bikeshed#254): ...`). That was my first pass at this (and I think I still prefer it), but it seemed too GitHub-centric. *shrug*

mikewest:
I went ahead and added support for the GitHub-specific syntax, as it feels like something I'll actually use.

Assuming the bot is happy, WDYT?

tabatkins:
I'm fine with specialized syntaxes if they're likely to be useful and they don't claim too much syntax space so we can add to them later.  This looks fine.

bikeshed/__init__.py
```diff
@@ -796,9 +796,21 @@ def decorateAutolink(doc, el, linkType, linkText):
...
     import hashlib
-    # Add an auto-genned and stable-against-changes-elsewhere id to all issues.
+    # Add an auto-genned and stable-against-changes-elsewhere id to all issues, and
+    # link to remote issues if possible:
     for el in findAll(".issue:not([id])", doc):
         el.set('id', "issue-"+hashContents(el))
+        remoteIssueID = el.get('data-remote-issue-id')
+        if remoteIssueID:
+            del el.attrib['data-remote-issue-id']
+            githubMatch = re.match(r"([^/]+)/([^#]+)#(\d+)", remoteIssueID)
```
tabatkins: This regex overclaims syntax space; it'll match on `Issue(Hahaha, this/looks like a github url#123456789 but it's not.)`.

I think we can get by `r"([\w-]+)/([\w-]+)#(\d+)$"`.

tabatkins: Or let's not be dogmatic about surrounding whitespace; `r"\s*([\w-]+)/([\w-]+)#(\d+)\s*$"

mikewest:
Strictified. WDYT?

bikeshed/__init__.py
```diff
@@ -796,9 +796,21 @@ def decorateAutolink(doc, el, linkType, linkText):
...
     import hashlib
-    # Add an auto-genned and stable-against-changes-elsewhere id to all issues.
+    # Add an auto-genned and stable-against-changes-elsewhere id to all issues, and
+    # link to remote issues if possible:
     for el in findAll(".issue:not([id])", doc):
         el.set('id', "issue-"+hashContents(el))
+        remoteIssueID = el.get('data-remote-issue-id')
+        if remoteIssueID:
+            del el.attrib['data-remote-issue-id']
+            githubMatch = re.match(r"\s*([\w\-_]+)/([^\w\-_]+)#(\d+)\s*$", remoteIssueID)
```
tabatkins: `r"\s*([\w\-_]+)/([^\w\-_]+)#(\d+)\s*$"` I think you messed up that middle one; you almost certainly don't mean to negate it.  Also, `_` is already part of `\w` so you can drop it, and you don't need to escape `-` if you put it last in the character class.

mikewest:
*facepalm* Thanks. I didn't even notice that the test had horrible output. :( Fixed in the most recent patch.

PR #377, opened 2015-03-10
Use an OrderedDict to maintain order for custom metadata.

beverloo:
Custom metadata in the generated specification should maintain the
order by which they were defined in the input .bs file.

### DIFF
--- a/bikeshed/MetadataManager.py
+++ b/bikeshed/MetadataManager.py
@@ -2,6 +2,7 @@
 from __future__ import division, unicode_literals
 import re
 import os
+from DefaultOrderedDict import DefaultOrderedDict
 from subprocess import check_output
 from collections import defaultdict
 from datetime import date, datetime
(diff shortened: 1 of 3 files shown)


bikeshed/MetadataManager.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, doc):
         self.issues = []
         self.issueTrackerTemplate = None
 
-        self.otherMetadata = defaultdict(list)
+        self.otherMetadata = OrderedDict()
```
tabatkins: I have a DefaultOrderedDict class, just `from DefaultOrderedDict import DefaultOrderedDict`.

(I really hate the "if empty" checks; they're a major source of bugs in my experience. defaultdict is a life-saver.)

beverloo: Done!

PR #385, opened 2015-04-12
add stacks to railroad diagrams

plinss:
Stacks work in any context, here's a nice test:
<pre class=railroad>
Stack:
    T: one
    T: two
    T: three
Stack:
    T: aaa
    T: bbb
Sequence:
    T: pre
    Stack:
        T: first
        T: second
    Stack:
        T: third
        T: fourth
    T: post
OneOrMore:
    Stack:
        T: 111
        T: 2222222
    T: ,
ZeroOrMore:
    Stack:
        T: 33333
        T: 4
    T: ;
Choice: 2
    T: first chice
    Stack:
        T: second
        T: choice
    Stack:
        T: third
        T: choice
    Stack:
        T: fourth
        T: choice
    T: fifth choice
</pre>

stacks

### DIFF
--- a/bikeshed/railroaddiagrams.py
+++ b/bikeshed/railroaddiagrams.py
@@ -32,6 +32,7 @@ class DiagramItem(object):
         self.children = [text] if text else []
         self.needsSpace = False
         self.raw = None
+        self.yAdvance = 0
 
     def format(self, x, y, width):
         raise NotImplementedError  # Virtual
@@ -110,8 +111,12 @@ class Diagram(DiagramItem):
         self.items = [Start()] + [wrapString(item) for item in items] + [End()]
         self.width = 1 + sum(item.width + (20 if item.needsSpace else 0)
                              for item in self.items)
-        self.up = max(item.up for item in self.items)
-        self.down = max(item.down for item in self.items)
+        self.up = 0
+        self.down = 0
+        for item in self.items:
+            self.up = max(self.up, item.up - self.yAdvance)
+            self.down = max(self.down, item.down + self.yAdvance)
+            self.yAdvance += item.yAdvance
         self.formatted = False
 
     def format(self, paddingTop=20, paddingRight=None, paddingBottom=None, paddingLeft=None):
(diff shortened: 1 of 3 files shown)


plinss:
Of course, the above is only to show that stacks work anywhere, the primary use is expected to be at the root, such as:

<pre class=railroad>
Stack:
    Sequence:
        T: terminal
        Optional:
            T: optional
        Choice:
            T: choice one
            T: choice two
            T: choice three
        OneOrMore:
            T: repeatable
            T: ,
    Sequence:
        N: non-terminal
        Choice: 2
            T: choice one
            T: choice two
            T: choice three
            T: choice four
            T: choice five
            T: choice six
        ZeroOrMore:
            T: more repeatable
            T: ,
</pre>

stacks2

tabatkins:
In the main repo I'm thinking I want to limit Stack to being the last element in a sequence.  If you don't, it's possible for things later in the sequence to collide with parts of the Stack, because the Stack can't take their heights into account when allotting space.  Also, anything following a Stack is pushed into the lower level; requiring them to nest in the Stack makes this a little more clear.

plinss:
I don't believe it's possible for anything to collide with a stack in my implementation. The stack takes the full width of its widest child (and centers the other children) and takes all the heights of its children into account (the first child can extend up, the others all extend below). Any following siblings of the stack are drawn to the right of the entire stack (offset vertically so their entrance path lines up with the exit path of the stack).

While using stacks in odd places can result in potentially ugly diagrams, I don't see any need or benefit to restricting their placement (it can be useful, for example, in stacking a single wide item inside a choice).

tabatkins:
Oh, so the following:

````
Stack:
    Sequence:
        T: terminal
        Optional:
            T: optional
        Choice:
            T: choice one
            T: choice two
            T: choice three
        OneOrMore:
            T: repeatable
            T: ,
    N: non-terminal
Choice: 2
    T: choice one
    T: choice two
    T: choice three
    T: choice four
    T: choice five
    T: choice six
ZeroOrMore:
    T: more repeatable
    T: ,
```
(same as your last diagram, but with the  Choice and ZeroOrMore promoted to siblings of the Stack)

would draw "non-terminal" all by itself on the last stack layer, with a bunch of empty line, then the Choice would start fully to the right of the Stack?

plinss:
yes

tabatkins:
K, merged.  Unfortunately, I now need to port this to railroad-diagrams, and get the JS and Python synced. ^_^

plinss:
Heh, yeah, I saw the PR there which inspired this (and it's useful for another project I'm using this for). I liked the idea but wanted a more flexible implementation.

PR #398, opened 2015-04-21
<code> is an opaque element.

mikewest:
This fixes some breakage for https://w3c.github.io/webappsec/specs/credentialmanagement/. I haven't dug into the details as to why, but I use `<code>...</code>` pretty frequently in the doc, so I can imagine it's wrapping things that would otherwise be processed?

### DIFF
--- a/bikeshed/htmlhelpers.py
+++ b/bikeshed/htmlhelpers.py
@@ -333,7 +333,7 @@ def isElement(node):
     return etree.iselement(node) and isinstance(node.tag, basestring)
 
 def isOpaqueElement(el):
-    return el.tag in ('pre', 'style', 'script')
+    return el.tag in ('pre', 'code', 'style', 'script')
 
 def fixTypography(text):
     # Replace straight aposes with curly quotes for possessives and contractions.


tabatkins:
I made `<code>` *not* an opaque element due to some DOM stuff literally a few days ago. ^_^

"opaque element"  is currently used solely to tell Bikeshed whether or not it should try to find markup shorthands (like `{{Foo}}`) in the element.  DOM had a lot of examples using `<code>` that wanted to link to methods with the IDL markup shorthand, so I relaxed the restrictions and removed `<code>` from the list.

If you're breaking, it's because you've got stuff that looks like markup shorthands in your <code>. I'll go download the spec and see what's happening.

tabatkins:
Ah, I see the issue.  Your private slot definitions are wrapped in a `<code>`, which previously kept them from being parsed as biblio refs.  That protection has been stripped away, so the biblio markup shorthand recognizer triggers, and transforms them into `[type]` (a single set of [] brackets).  This triggers two errors - it's not a valid biblio ref, and the definition is now wrong so your autolinks are breaking.

Solution is to escape the name: `<dfn attribute>\[[type]]</dfn>`, etc.

mikewest:
Hrm. I was using this a lot. Content Security Policy has source expressions that match the various kinds of CSS shorthands that Bikeshed supports. Escaping all the usage is a bit annoying.

Could we add a mechanism whereby a document could opt-out of certain shorthand?

mikewest:
And, after going through a zillion lines, it doesn't look like escaping works for `'this-kind-of-thing'`. :(

tabatkins:
You want `Markup Shorthands: css off` in your metadata.

tabatkins:
And yeah, I havent' covered all the shorthands with escapes yet, sorry. ;_;

PR #400, opened 2015-04-24
Echidna doesn't like commas after the document status.

mikewest:
See https://twitter.com/robinberjon/status/591543288336932864.


mikewest:
Huzzah, w3c/specberus#188 fixed it. :)

PR #405, opened 2015-05-07
Remove the "manually provide IDs for your headings" warning

foolip:
A lot of the time, the ID should match the text. Adding one manually
in this case is a good way to later get your IDs out of sync with
the text to varying degrees.

### DIFF
--- a/bikeshed/__init__.py
+++ b/bikeshed/__init__.py
@@ -263,18 +263,13 @@ def resetHeadings(doc, headings):
         appendChild(header, content)
 
 def addHeadingIds(doc, headings):
-    neededIds = set()
     for header in headings:
         if header.get('id') is None:
-            neededIds.add(header)
             header.set('id', simplifyText(textContent(find(".content", header)), convertDashes=True))
         if header.get("oldids"):
             oldIDs = [h.strip() for h in header.get("oldids").strip().split(",")]
             for oldID in oldIDs:
                 appendChild(header, E.span({"id":oldID}))
-    if len(neededIds) > 0:
-        warn("You should manually provide IDs for your headings:\n{0}",
-            "\n".join("  "+outerHTML(el) for el in neededIds))
 
 def determineHeadingLevels(doc, headings):
     headerLevel = [0,0,0,0,0]


foolip:
Alternatively, how about a way to suppress the warning? Maybe `<h1 id>bla</h1>` as a signal that the generated ID is the right one?

CC @richtr

tabatkins:
Generating one manually is a good way to have links break just because you slightly reworded the heading.  The warning was put there intentionally. ^_^

(If you do change the text a lot and want to change the id to match, put the old id into an `oldids` attribute on the heading. That'll preserve links.)

foolip:
Definitions have auto-generated IDs with no warnings, is the idea that those are less likely to change?

To illustrate what I think [the problem](https://github.com/whatwg/mediasession/commit/7cf9ffc9d3536b162574e0f56be1f13b2553816f#diff-c850468e00d00d6d586eb12bf8399017R246) is:
```
<h3 id="extensions-to-the-htmlmediaelement-interface">Extensions to the
{{HTMLMediaElement}} interface</h3>
```

Apart from being very verbose, manually maintained IDs have their own risk, as you've seen:
https://github.com/w3c/web-animations/issues/88
https://github.com/w3c/web-animations/issues/89

If you think the warning has positive value on average, how about a way to silence it?

tabatkins:
> Definitions have auto-generated IDs with no warnings, is the idea that those are less likely to change?

Precisely.  Headings change regularly, definitions change much more rarely.

> Apart from being very verbose

That's entirely up to the spec author.  Nothing preventing you from generating short ids. Heck, just make them "a", "b", etc if you want.

On the other hand, it's easy for auto-genned IDs to be *very* verbose for headings, particularly if the heading is well-written according to the CSSWG style - "Flex Flow Direction: the flex-direction property" would gen a *much* longer id than the manually specified one #flex-direction-property.

It's of course up to the spec author to make sure their IDs still more-or-less match up with the heading text.  The ID and text are *right next to each other*, so this isn't hard.

> If you think the warning has positive value on average, how about a way to silence it?

That defeats the entire point; if you silence it, you'll just regularly break links as you tweak heading text.  Put an `id="a"` or `id="b"` on it if you don't care what the ID is and just want something short that doesn't need maintenance.

fantasai:
That example you posted illustrates the problem adequately well. I would code it as
<h3 id="htmlmediaelement-interface-extensions">
Extensions to the HTMLMediaElement Interface</h3>
Not only does this make the URL more compact (easier to copy/paste without wrapping errors, easier to see the whole thing in the URL bar, etc), when I later decide that the title should be “HTMLMediaLement Interface Extensions” or “Additions to the HTMLMediaElement Interface” or something else, nobody's links get broken. As Tab points out, this is much less likely to happen with terminology links, because terms don't have so many opportunities for rewording.

foolip:
In the example I gave, the autogenerated ID is precisely what I want, like several other specs:
http://dev.w3.org/csswg/cssom/#extensions-to-the-document-interface
http://dev.w3.org/csswg/cssom-view/#extensions-to-the-htmlimageelement-interface
http://w3c.github.io/selection-api/#extensions-to-document-interface
http://w3c.github.io/webcomponents/spec/shadow/#extensions-to-element-interface
http://w3c.github.io/web-animations/#extensions-to-the-document-interface

(These links end up in Blink's IDL files and I see no reason to deviate from the pattern.)

In short, I do care what the IDs are, I almost always want them to match the (usually short) text, and already understood before reading the warning that changing the text would change the autogenerated ID. I guess the style of section titles in CSS specs changes the balance of pros and cons.

Other than removing the warning, is there *any* change to use the autogenerated IDs without warnings that would be acceptable?

tabatkins:
No, because they break links. Even the example you gave of manually-set IDs being problematic (in the WebAnim spec) are still good, because they're stable and links to them continue to work.  If the IDs had been auto-generated, any links would have broken as soon as the interface name changed, and most likely the editor wouldn't have realized the problem.  As it is, the mismatch was obvious (though the editor still didn't notice, or chose to keep the IDs stable), and with the use of `oldids` you can have both a correct ID and maintain old links.

Auto-generating IDs for headings was rejected for CommonMark for the same reason. ^_^

Just specify your ID; it's not a big deal, and it's far better for long-term usability.  (This is specifically why the CSSWG markup style puts the opening tag of a heading on a separate line from the text; so a well-written ID and heading text don't look "too long" in the source.)

foolip:
One can maintain old links with `oldids` with or without autogenerated IDs and warnings about them.

We're not going to agree, but thanks anyway for considering this.

tabatkins:
I hope you agree that one is *much* less likely to remember to add a value to `oldids` if the ID is being auto-generated.

foolip:
Sure, that seems less likely. Maybe you should advertise `oldids` with "NOTE: Heading IDs were generated, please manually provide IDs or use oldids when changing the headings."

PR #416, opened 2015-06-04
add w3c-status metadata

plinss:
also fixed csswg-work-status for EDs


plinss:
Another item I'd like to scrape form the specs is the next transition date, i.e. the minimum date required for transitioning to CR or PR... thoughts?

tabatkins:
That's the `Deadline` metadata, so I can plumb that trivially.

PR #422, opened 2015-06-29
This test doesn't produce the output I'd expect for an optional dictionary.

mikewest:
Not sure if this is a bug in my understanding of IDL (see below), the IDL parser, or Bikeshed. Regardless, this cod (and the optional dictionary member on line 24 in particular) doesn't produce the output I'd expect.

https://heycam.github.io/webidl/#dfn-optional-argument-default-value says:

> If the type of an argument is a dictionary type or a union type that has a dictionary type as one of its flattened member types, and that dictionary type and its ancestors have no required members, and the argument is either the final argument or is followed only by optional arguments, then the argument must be specified as optional. Such arguments are always considered to have a default value of an empty dictionary, unless otherwise specified.

My reading of that is that single-argument methods with dictionary arguments should mark the arguments as optional. *shrug* I Have No Idea What I'm Doing, though.

### DIFF
--- a/tests/idl001.bs
+++ b/tests/idl001.bs
@@ -1,0 +1,26 @@
+<h1>Foo</h1>
+
+<pre class=metadata>
+Group: test
+Shortname: foo
+Level: 1
+Status: ED
+ED: http://example.com/foo
+Abstract: A document with IDL content.
+Editor: Example Editor
+Date: 1970-01-01
+</pre>
+
+<pre class="idl">
+    dictionary ExcitingDictionary {
+        boolean thingOne;
+        boolean thingTwo;
+    };
+
+    interface AnInterface {
+        boolean aMethod(boolean anArg);
+        boolean aMethodToo(optional boolean anArgToo);
+        boolean aMethodThree(ExampleDictionary dict);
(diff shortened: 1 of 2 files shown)


tabatkins:
What are you expecting?  That the IDL parser should reject that IDL with an error because it's not marked as optional?  I don't currently (nor do I particularly plan to) actually validate the IDL beyond a base syntactic level.

mikewest:
If line 24 is present, it produces a `FATAL ERROR: No 'idl-name' refs found for 'ExampleDictionary'.` (see line 107 of https://travis-ci.org/tabatkins/bikeshed/builds/68790418). :)

I was seeing some issues with the argument name not being defined as well, but I'm not sure if they're still present.

tabatkins:
I expect line 23 to cause that issue as well, *because ExampleDictionary isn't defined*.  The dictionary in your example is called ExcitingDictionary.

Am I missing something, or is this just a matter of typo-blindness? ^_^

tabatkins:
Ping!

mikewest:
I dropped this on the floor long enough that I have no idea what the complaint was. If it crops up again, I'll file a new PR. :)

PR #426, opened 2015-07-02
Strip trailing whitespace when injecting a header's link.

mikewest:
@timeless noted in https://lists.w3.org/Archives/Public/public-webappsec/2015Jun/0066.html that many of the header-based links have trailing whitespace. This patch pokes at that a bit. WDYT?


tabatkins:
Wanna just do a full `.strip()`, to be safe? No reason to keep any preceding whitespace, either.

mikewest:
SGTM.

PR #433, opened 2015-07-30
Add a note that MacPorts shouldn't be needed on newish OSX.

miketaylr:
I'm on 10.10 and just hopped straight to cloning and pip without issue.


miketaylr:
:heart:

PR #450, opened 2015-08-04
Add boilerplate for the Web Bluetooth CG.

jyasskin:


tabatkins:
The logo.include files are raw HTML, not stylesheets.  If you need special styles, put them in your header.

jyasskin:
I need 2 different special styles for CG-DRAFT vs CG-FINAL, and I'd rather not duplicate the whole header. I could add a logo-style.include for style-based logos like CGs use. I could also link to https://www.w3.org/community/src/css/spec/[STATUS].css like header-webappsec does, but that pulls in some other styles that don't look like what we want.

tabatkins:
Maybe just wrap it up in a `<style>`, then?  As long as it's HTML it's fine, and Bikeshed will move it into the `<head>` where it belongs automatically.

jyasskin:
How's this?

PR #463, opened 2015-08-22
Add boilerplate for Houdini taskforce.

dbaron:
This copies data from the CSSWG templates.  I didn't check the diffs
very carefully other than changing www-style to public-houdini
throughout, and changing the github repository in
defaults-houdini.include.


tabatkins:
Gave it a quick lookover, but nothing serious.  If there are any problems we'll notice them at some point and fix them then.

Issue #477, opened 2015-09-03
Differences in GitHub issue handling when running Bikeshed locally


foolip:
I have switched back to using the web service for Media Session:
https://github.com/whatwg/mediasession/commit/5a9999d1cc17e2b5bdfd4130504765d1a454e526

However, there are still differences I can't explain in the diff:

```
-     <dd><a href="https://github.com/whatwg/mediasession/issues/">GitHub</a>
```

and

```
   <div style="counter-reset:issue">
-   <div class="issue"> The interruption may not come from another media session at all. <a href="https://github.com/whatwg/mediasession/issues/100">&lt;https://github.com/whatwg/mediasession/issues/100></a><a href="#issue-8519f2d2"> ↵ </a></div>
-   <div class="issue"> The interruption may not come from another media session at all. <a href="https://github.com/whatwg/mediasession/issues/100">&lt;https://github.com/whatwg/mediasession/issues/100></a><a href="#issue-8519f2d20"> ↵ </a></div></div>
+   <div class="issue"> The interruption may not come from another media session at all.<a href="#issue-8519f2d2"> ↵ </a></div>
+   <div class="issue"> The interruption may not come from another media session at all.<a href="#issue-8519f2d20"> ↵ </a></div>
+  </div>
```

There's also a similar difference in the place where the issues are mentioned.

tabatkins:
If you're using the web service, Bikeshed can't auto-detect that you're in a GH repo.  You can tell it manually by adding a `Repository: whatwg/mediasession/` line to your metadata.

foolip:
Thanks!

foolip:
Actually, is there any way to suppress the generation of the Issue Tracking link? It really wouldn't add anything to https://mediasession.spec.whatwg.org/ so if I could not have it generated locally, that would be even better.

domenic:
I think this is what does the trick https://github.com/whatwg/streams/blob/master/index.bs#L9

tabatkins:
Yup, what domenic said.

I'm starting to work on a documentation rewrite, and dealing with this part specifically will be getting a much better treatment. >_<

PR #478, opened 2015-09-04
Add missing self argument to Repository.formatIssueUrl()

foolip:
Triggered by adding 'Repository: whatwg/mediasession/' to the config as per
https://github.com/tabatkins/bikeshed/issues/477#issuecomment-137526035

To be recognized as GitHub a full URL is needed, it seems.


tabatkins:
...how did no one run into this already?

tabatkins:
And no, omitting the trailling slash should work.

Issue #475, opened 2015-09-02
Convert `text` to <code>text</code>


jyasskin:
I write lots of JS code in specs, and it'd be nice to have a shorthand for it.

tabatkins:
I plan to do this as part of a more general "support the inline Markdown patterns" edit.  It's not hard to do (especially now that the spec author can control which sets of markup shorthands they want to use), I just haven't gotten to it yet.

mikewest:
I want this too. Like, lots. :) So, :+1::100::+1:.

tabatkins:
Okay, added.  To turn it on you have to activate inline markdown via `Markup Shorthands: markdown on` (you can put it in your WG's default.include if you want).  I haven't implemented the rest of inline markdown yet, but they'll come shortly.

As far as I can tell it handles everything correctly - multiple numbers of backticks, contained backticks, etc.  Stuff inside is HTML-escaped, and per CommonMark, whitespace is stripped from the start/end and collapsed internally.  Let me know if you run into any bugs.

tabatkins:
Oh! This does *not* handle intermixing with HTML correctly.  If you have a backtick in an attribute value, it'll trigger backtick handling and mangle your HTML.  I have to do some complicated things to fix this, so I haven't done it yet, but I know how to do it.

PR #490, opened 2015-09-22
Fix the backtick markdown shortcut.

mikewest:
The existing code built up a stack of code block replacements. It
should probably have been a queue. This patch takes the simplest
possible approach by reversing the list before processing the
replacements.

Refixes tabatkins/bikeshed#475.


mikewest:
Hi @tabatkins! I think your backtick processing code was backwards. :)

tabatkins:
Oh my gosh, I can't believe I mixed up my stacks and my queues. I'm a dumb dummy.

Issue #492, opened 2015-09-24
Inlining GitHub Issues hits API rate limit


tobie:
One workaround is to use the API to [list issues of a repository](https://developer.github.com/v3/issues/#list-issues-for-a-repository) instead. Make sure to fetch them all using the data provided by the [Link headers](https://developer.github.com/v3/#link-header). This should limit the number of requests to 3-4 for a large repo, instead of _n_ as is currently the case.

Another (better) workaround is to allow passing credentials to bikeshed which should boost the number of authorized requests from 60 per hour to 5000. See https://developer.github.com/v3/#rate-limiting. 

---

* [x] Cache responses locally, and use conditional queries to see if anything changed <https://developer.github.com/v3/#conditional-requests>
* [ ] Use pagination to make fewer requests; construct paging from the requested issues <https://developer.github.com/v3/#pagination>
* [x] Let people use auth to increase their own API limits <https://developer.github.com/v3/#rate-limiting>

tabatkins:
Okay, I've editted in a list of things I need to do. I'm being terrible naive in my treatment right now.

tabatkins:
All right, caching and conditional responses are now in use, so hitting rate limits should be *much* rarer now; you'll only run into it if you're processing for the first time and use a bunch (> 60) of inline issues, and then it just means you have to wait an hour for your limit to refresh and Bikeshed will pick up where it left off.  (Bikeshed emits an informative error message for this case.)  Or get an OAuth token once, so it can blast thru your issues list; unless more than 60 issues have changed since the last time you ran Bikeshed, you won't need that token again.

Paging turned out to be a dud; I can't reliably predict what pages the issues will be on unless I slurp down *all* issues including closed ones, and (a) I don't think I want to speculatively cache every issue in your whole repo, and (b) I'm not sure how conditional responses work in that case.  Making individual requests per issue isn't a big deal now, so I'm going to consider this issue closed.

PR #494, opened 2015-09-26
Add support for passing GH tokens via the command line.

tobie:
Partially addresses #492.

### DIFF
--- a/bikeshed/__init__.py
+++ b/bikeshed/__init__.py
@@ -2561,12 +2561,15 @@ def inlineRemoteIssues(doc):
         except urllib2.HTTPError as err:
             if doc.token and err.code == 401:
                 die("Unauthorized Access to GitHub's API. There might be an issue with your token.")
-            elif doc.token is None and err.code == 403:
-                die("You've reached GitHub API's rate limit. Please provide an auth token.")
-            else:
-                die("{0}", err)
+                return
+            if doc.token is None and err.code == 403:
+                die("You've reached GitHub API's rate limit for unauthenticated requests.\nTo increase it, please provide an auth token. Tokens can be generated from https://github.com/settings/tokens.")
+                return
+            die("Error inlining GitHub issues:\n{0}", err)
+            return
         except Exception as err:
-            die("{0}", err)
+            die("Error inlining GitHub issues:\n{0}", err)
+            return
 
 def addNoteHeaders(doc):
     # Finds <foo heading="bar"> and turns it into a marker-heading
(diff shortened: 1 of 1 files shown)


bikeshed/__init__.py
```diff
@@ -2551,8 +2558,15 @@ def inlineRemoteIssues(doc):
                 if el.tag == "p":
                     el.tag = "div"
                 addClass(el, "no-marker")
-        except:
-            pass
+        except urllib2.HTTPError as err:
+            if doc.token and err.code == 401:
+                die("Unauthorized Access to GitHub's API. There might be an issue with your token.")
+            elif doc.token is None and err.code == 403:
+                die("You've reached GitHub API's rate limit. Please provide an auth token.")
```
tabatkins: Can you provide a URL here?

bikeshed/__init__.py
```diff
@@ -2551,8 +2558,15 @@ def inlineRemoteIssues(doc):
...
+            else:
+                die("{0}", err)
```
tabatkins: Here and the next die(), please supply an intro message, then stringify the error message onto the next line.  Like, `"Error inlining GitHub issues:\n{0}"`.

bikeshed/__init__.py
```diff
@@ -2551,8 +2558,15 @@ def inlineRemoteIssues(doc):
                 if el.tag == "p":
                     el.tag = "div"
                 addClass(el, "no-marker")
-        except:
-            pass
+        except urllib2.HTTPError as err:
+            if doc.token and err.code == 401:
+                die("Unauthorized Access to GitHub's API. There might be an issue with your token.")
```
tabatkins: Please put a `continue` or `break` after each of these die()s, as appropriate.  There's nothing after this block *now*, but there might be in the future, so it's best to make sure the die() properly stops working on the iteration as soon as it find an error.

tobie: Duh. `Die` sounded so imperious I sort of wrongly assumed it terminated the process. I feel like auth/rate limit should both stop the iteration and return from `inlineRemoteIssues`. I imagine the correct way to do that in python is just to return? Right?

tobie: Added `return` statements were appropriate. LMK if that's OK for you.

tabatkins: die() *does* terminate the process *by default*, but it's common and expected to sometimes run bikeshed with the `--force` option, which prevents that.  I try my best to recover gracefully from any fatal errors.

`return` is too strong here; you only need to exit the loop.  If people are embedding inline issues from multiple types of repositories, for example (which isn't supported now, but could be in the future), the GH issues failing shouldn't prevent any following types of issues from working.

I think the 401 and 403 errors should use `break`, and the other two probably just a `continue`, in case it's a transient issue that further requests won't suffer from.

tobie: Despite this note still showing up, your comment has been addressed. :)

PR #497, opened 2015-10-01
Add boilerplate for the geolocation WG. 

mkruisselbrink:
Mostly copied from the webappsec templates.


tabatkins:
If you're not making any alterations to the CSSWG stylesheet, you can omit that; Bikeshed falls back to the generic includes when it can't find a more specific one, so you'll just use the `stylesheet.include` file.  I actually maintain that one to be consistent with the CSSWG, so it'll be updated.

If you *are* making changes, but are using most of the CSSWG stylesheet, it's probably still best to just rely on the default stylesheet.include and add your own styles after it in your header inclusion.

mkruisselbrink:
Actually I copied that file from the webappsec specs. But that apparently is just an older version of the stylesheet that is included in bikeshed already? Then there probably is no reason for me to use that stylesheet, so I've deleted it.

PR #499, opened 2015-10-03
Make DAP's conformance section overwritable.

tobie:


tabatkins:
This'll work, and I'll merge it, I'm just curious what your use-case is for having different conformance requirements per spec.

tobie:
It's a special case for the Generic Sensor API. All the examples are made up. And so I needed to add a [little text](https://github.com/w3c/sensors/blob/101f11e/index.bs#L803-L809) t to explain that.

tabatkins:
Huh, okay.  I'd've just put that in the intro or something, but whatever floats your boat.

(Boilerplate is just there to CYA and to point to for complete newbs; it's read by virtually no one, and so anything put there might as well be invisible.)

tobie:
Yeah. Maybe you're right and it belongs elsewhere. It was requested by the chair of one of the WG which might or might not decide to use the sensor api, so it *is* sort of CYA-type text.

PR #503, opened 2015-10-14
Fold whitespace when processing `<var>`s

mikewest:
`|this is a var|` is the same as `|this\nis a var|` in an HTML document. This
patch teaches the singular-var-checker about this interesting property.


mikewest:
WDYT, @tabatkins?

tabatkins:
I disagree with y'all's silly "linebreak at an arbitrary column" ways, but I'm fine with this patch. ^_^

PR #505, opened 2015-10-14
Accept 'algorithm' as a bare attribute.

mikewest:
If h*@algorithm is present, but empty, the header's text will be used as the algorithm name.

This will make integrating it into my various specs a lot simpler and less repetitive. :)


tabatkins:
Good idea!

Issue #513, opened 2015-10-30
Bikeshed should require security and privacy considerations sections.


mikewest:
As discussed in https://w3ctag.github.io/security-questionnaire/#considerations, it would be lovely if Bikeshed would fail to compile specifications that don't contain a header labeled "Security Considerations" and a header labeled "Privacy Considerations".

WDYT, @tabatkins?

/cc @npdoty @mnot

tabatkins:
I'm not down with requiring this for all Bikeshedded documents; Bikeshed is used outside the W3C, and can be used for non-spec things (I'm doing so privately), both of which don't need this.

But I *am* happy to maintain a list of groups that are W3C, and trigger the check/warning whenever a spec is generated for one of those groups.

mikewest:
Ok. What do you think about the approach in https://github.com/tabatkins/bikeshed/pull/514?

npdoty:
Should we also check for a "Privacy and Security Considerations" section? It seems like some groups are combining them, which seems like a reasonable practice we should support. Or was there a particular reason to prompt for separate sections entitled "Privacy Considerations" and "Security Considerations"?

mikewest:
I don't have strong feelings about it. I'd be happy to support both in one section if that's a common practice.

tabatkins:
Done.

PR #514, opened 2015-10-30
Warn if security/privacy considerations sections are not present.

mikewest:
Closes tabatkins/bikeshed#513.

### DIFF
--- a/bikeshed/MetadataManager.py
+++ b/bikeshed/MetadataManager.py
@@ -19,6 +19,10 @@ class MetadataManager:
             return "{0}-{1}".format(self.shortname, self.level)
         return self.shortname
 
+    @property
+    def groupIsW3C(self):
+        return self.group in config.groupsInW3C
+
     def __init__(self, doc):
         self.doc = doc
         self.hasMetadata = False
--- a/bikeshed/__init__.py
+++ b/bikeshed/__init__.py
@@ -280,6 +280,8 @@ def processHeadings(doc, scope="doc"):
     addHeadingBonuses(doc, headings)
     for el in headings:
         addClass(el, 'settled')
+    if scope == "all" and doc.md.groupIsW3C:
+        checkPrivacySecurityHeadings(findAll(".heading", doc))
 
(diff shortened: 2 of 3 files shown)


mikewest:
Since you explicitly asked for Bikeshedding during your time off, @tabatkins, ping on this? :)

bikeshed/MetadataManager.py
```diff
@@ -19,6 +19,11 @@ def vshortname(self):
             return "{0}-{1}".format(self.shortname, self.level)
         return self.shortname
 
+    @property
+    def groupIsW3C(self):
+        return self.group in ["csswg", "dap", "fxtf", "geolocation", "houdini", "ricg", "svg",
+                              "texttracks", "web-bluetooth-cg", "webappsec", "webspecs", "whatwg"]
+
```
tabatkins: I put data like this in the config.py file. Make a frozenset there with these names. I like the function still, just want the data moved. ^_^

bikeshed/__init__.py
```diff
@@ -308,6 +314,13 @@ def addHeadingIds(doc, headings):
     if len(neededIds) > 0:
         warn("You should manually provide IDs for your headings:\n{0}",
             "\n".join("  "+outerHTML(el) for el in neededIds))
+    if doc.md.groupIsW3C:
+        if not hadSecurity and not hadPrivacy:
+            warn("This specification has neither a 'Security Considerations' nor a 'Privacy Considerations' section. Please consider adding both.")
```
tabatkins: Warning doesn't tell you the magic ID it's looking for.  Either say "a #security-considerations section"/etc, or change the check to look for text content rather than ID.

mikewest: Changed to look at the header's text instead.

bikeshed/MetadataManager.py
mikewest: Got it. Done.

mikewest:
Rebased and fixed up based on your feedback. WDYT? :)

Issue #538, opened 2015-12-13
Markdown horizontal rules


arronei:
It would be great if the horizontal rule markdown functionality was supported.

\***
\---
\___

PR #539, opened 2015-12-13
Add horizontal rule

arronei:
Adding the functionality for markdown horizontal rules.

Fixes: #538

### DIFF
--- a/bikeshed/markdown.py
+++ b/bikeshed/markdown.py
@@ -68,7 +68,7 @@ def tokenizeLines(lines, numSpacesForIndentation, features=None, opaqueElements=
 			if match:
 				token['id'] = match.group(1)
 		elif re.match(r"([*_-])\s*\1\s*\1", line):
-			token = {'type':'rule', 'text': "", 'raw': rawline}
+			token = {'type':'rule', 'raw': rawline}
 		elif re.match(r"-?\d+\.\s", line):
 			match = re.match(r"(-?\d+)\.\s+(.*)", line)
 			token = {'type':'numbered', 'text': match.group(2), 'raw':rawline, 'num': int(match.group(1))}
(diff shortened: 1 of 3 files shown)


bikeshed/markdown.py
```diff
@@ -67,6 +67,8 @@ def tokenizeLines(lines, numSpacesForIndentation, features=None, opaqueElements=
 			match = re.search(r"\{#([^ }]+)\}\s*$", line)
 			if match:
 				token['id'] = match.group(1)
+		elif re.match(r"([*_-])\s*\1\s*\1", line):
```
tabatkins: Need a `(\s*\1)*$` at the end to ensure that the line doesn't have anything else on it, per CommonMark.

bikeshed/markdown.py
```diff
@@ -217,6 +221,11 @@ def parseMultiLineHeading(stream):
 	stream.advance(2)
 	return lines
 
+def parseHorizontalRule(stream):
+	lines =  ["<hr />\n"]
```
tabatkins: Just `<hr>`, please.  (I parse it as HTML anyway.)

tabatkins:
Gave some nits; fix and I'll merge.

And lolwtf how did I get that crazy indentation.  Goddammit, Python, THIS IS WHY YOU SHOULD HAVE STANDARDIZED ON TABS.

arronei:
I'm not seeing this error locally. The regex is valid from what I can tell. Any ideas what I'm missing?

tabatkins:
Yes, the regex is valid. But it didn't fully match the CommonMark spec.

arronei:
Got it. I'll change back to what I have with the tweak you noted.

tabatkins:
All right, I tweaked the regex to make it not fail (which is weird, it should have worked), and fixed one line to match CommonMark more closely. All merged, thanks!

PR #540, opened 2015-12-21
Editorial change

FremyCompany:
Self explanatory


tabatkins:
Thanks! (I really need to figure out why that one particular test is failing - I suspect file locations are the reason.)

PR #541, opened 2015-12-21
Fixed propdef, at least according to my understanding

FremyCompany:


FremyCompany:
I should maybe have used zero-width spaces. That's what I ended up doing in my spec.

tabatkins:
Nah, this isn't right. CSS grammar nonterminals are done with `<<foo>>`.

FremyCompany:
I don't know about that. The sample in your sample isn't right either. I am not sure what to use, but I am sure what the sample says to use does not compile ;-)

FremyCompany:
Could you please check this works?
```
    	<pre class='propdef'>  
    	Name: var-*  
    	Values: [ <value> | <CDO> | <CDC> ]  
```

tabatkins:
Yeah, what I have there is definitely wrong.  It should be changed to either `[ <value> | <CDO> | <CDC> ] ` (which'll trigger linking failures if used in a real spec, since those nonterminals don't exist, but it's the general way to do it) or `[ &lt;value> | &lt;CDO> | &lt;CDC> ]`.  I prefer the former, so it generates links correctly when people use their values.  Maybe change it to the current value definition of `<<declaration-value>>`?

tabatkins:
Switched it to a different property that links more easily anyway.

PR #546, opened 2016-01-07
Adding types and dfn handling

arronei:
* Add new group "html"
* Adding new types to types list
* Add "for" handling for all dfns


arronei:
Opened issue #547 on me to finish cleanup in the future.

PR #567, opened 2016-01-19
linkTextVariations should kick in only if no direct match is found

FremyCompany:
Proposed patch for ReferenceManager.
I didn't test if that works, but I remember you have a test suite, so that should do the trick :D

### DIFF
--- a/bikeshed/ReferenceManager.py
+++ b/bikeshed/ReferenceManager.py
@@ -226,16 +226,20 @@ class ReferenceManager(object):
             return None
 
         # Local refs always get precedence, no matter what.
-        localRefs = self.getLocalRef(linkType, text, linkFor, linkForHint, el)
+        # At first let's try with "exact=true"...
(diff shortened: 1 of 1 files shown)


tabatkins:
Nah, correct fix was in the core search algorithm - it shouldn't even try to find variations if an exact match hit. Fixed now.

FremyCompany:
LGTM. Thanks.

annevk:
I think this commit caused a ton of breakage all over.

annevk:
E.g., try building DOM.

tabatkins:
@FremyCompany So my patch for this was wrong - it meant that if something was squatting on your exact term but wasn't good enough to link by default, and a variation would have grabbed the right term, you'd get a link failure.

I'll have to rewrite this to be closer to your originally suggested "just do it twice" code, so I can go through the entire matching rigmarole with the exact term first, and then fail over to using variations if that failed.  For now, I've reverted it because it was breaking several other specs.

PR #589, opened 2016-02-11
Master

garykac:
Add support for 'uievents' W3C group


bikeshed/config.py
```diff
@@ -42,7 +42,7 @@
 
 groupsInW3C = frozenset(["csswg", "dap", "fxtf", "geolocation", "houdini",
                          "html","ricg", "svg", "texttracks", "web-bluetooth-cg",
-                         "webappsec", "webspecs", "whatwg"])
```
tabatkins: Why did you remove webappsec?

garykac: That was not intentional. It should only be adding support for uievents.

PR #594, opened 2016-02-12
Made the ref error more meaningful

plehegar:


tabatkins:
This assumes that the 'spec' attribute was passed, which is often not true.  You need to keep the old line, and decide between the two based on whether 'spec' is `None`.

tabatkins:
And I've gone ahead and fixed this, adding the line to both of the failure reasons that happen after "spec", in case the spec filtering contributed to those failures.

Issue #595, opened 2016-02-12
Crash when trying to define a method with parameters in <pre class=anchors>


mkruisselbrink:
If I try to define method in a `<pre class=anchors>` block, something like this:
```
<pre class=anchors>
spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
    type: interface; text: GlobalFetch; url: dom-global-fetch
    type: method; text: fetch(input, init); for: GlobalFetch; url: dom-global-fetch
</pre>
```

bikeshed crashes with the following stacktrace:
```
Traceback (most recent call last):
  File "/usr/local/bin/bikeshed", line 9, in <module>
    load_entry_point('Bikeshed==0.0.0', 'console_scripts', 'bikeshed')()
  File "/usr/local/google/home/mek/specs/bikeshed/bikeshed/__init__.py", line 177, in main
    doc.preprocess()
  File "/usr/local/google/home/mek/specs/bikeshed/bikeshed/__init__.py", line 326, in preprocess
    self.lines = datablocks.transformDataBlocks(self, self.lines)
  File "/usr/local/google/home/mek/specs/bikeshed/bikeshed/datablocks.py", line 81, in transformDataBlocks
    doc=doc)
  File "/usr/local/google/home/mek/specs/bikeshed/bikeshed/datablocks.py", line 377, in transformAnchors
    return processAnchors(anchors, doc)
  File "/usr/local/google/home/mek/specs/bikeshed/bikeshed/datablocks.py", line 415, in processAnchors
    doc.refs.methods[arglessName].append(anchor['text'][0])
AttributeError: 'dict' object has no attribute 'append'
```

PR #596, opened 2016-02-12
Fix crash when defining a method in another spec.

mkruisselbrink:
This seems to fix the crash in issue #595


mkruisselbrink:
@tabatkins: ping, does this look like the correct fix?

tabatkins:
Yup, sorry for the delay.

PR #597, opened 2016-02-12
Add boilerplate for the web platform working group.

mkruisselbrink:
Also exposes mailing list information via macros, as it seemed silly to
duplicate that info in defaults.include and status.include.


bikeshed/include/status-webplatform-ED.include
```diff
@@ -0,0 +1,38 @@
...
+  <a href="mailto:[MAILINGLIST]?Subject=%5B[SHORTNAME]%5D%20PUT%20SUBJECT%20HERE">[MAILINGLIST]</a>
+  (see <a href="http://www.w3.org/Mail/Request">instructions</a>)
+  is preferred for discussion of this specification.
+  When sending e-mail,
+  please put the text “[SHORTNAME]” in the subject,
+  preferably like this:
+  “[<!---->[SHORTNAME]] <em>…summary of comment…</em>”
+
+<p>
+  This document was produced by the
```
mkruisselbrink: It seems kind of silly that every working group will more or less have to end up with their own 5 or so copies of status.include files. So given that at this point this paragraph is pretty much the only thing that would actually differ between working groups (at least for the csswg, dap, fxtf, houdini and webappsec templates), would it make sense to have a separate "procuded-by.include" and have that vary for working groups, and just have generic status--ED.include etc files for the rest of the shared boilerplate?

tabatkins: Yeah, that makes sense. I'll work on the separation today.

PR #643, opened 2016-03-22
Cleanup DAP styles.

tobie:


tabatkins:
Use `[W3C-STYLESHEET-URL]`, plz.  Otherwise you'll break on certain statuses, and on certain dates.

PR #644, opened 2016-03-24
Clean up HTML includes

arronei:
* Changed copyright to be permissive to match webplatform
* Changed style linking to link to correct styles and not be a 404
* Updated data-type value to trim whitespace


tabatkins:
Can you separate out the data-type fix from the boilerplate cleanup?  Also, use `unicode()`, not `str()`.

PR #663, opened 2016-04-12
Update HTML include files

arronei:
* Removed broken links
* Use the default styles from stylesheet.include
* Correct license information


bikeshed/include/html/copyright.include
```diff
@@ -1 +1 @@
...
+<a href="http://www.w3.org/Consortium/Legal/ipr-notice#Copyright">Copyright</a> © [YEAR] <a href="http://www.w3.org/"><abbr title="World Wide Web Consortium">W3C</abbr></a><sup>®</sup> (<a href="http://www.csail.mit.edu/"><abbr title="Massachusetts Institute of Technology">MIT</abbr></a>, <a href="http://www.ercim.eu/"><abbr title="European Research Consortium for Informatics and Mathematics">ERCIM</abbr></a>, <a href="http://www.keio.ac.jp/">Keio</a>, <a href="http://ev.buaa.edu.cn/">Beihang</a>). W3C <a href="http://www.w3.org/Consortium/Legal/ipr-notice#Legal_Disclaimer">liability</a>, <a href="http://www.w3.org/Consortium/Legal/ipr-notice#W3C_Trademarks">trademark</a> and <a rel="license" href="http://www.w3.org/Consortium/Legal/2015/copyright-software-and-document">permissive document license</a> rules apply.
```
tabatkins: Can you revert this?  All the include files have a trailing newline; no need to churn the history.

bikeshed/include/html/header.include
```diff
@@ -2,13 +2,11 @@
...
 <head>
   <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
-  <title>[TITLE] [LEVEL]</title>
   <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
   <meta name="w3c-status" content="[STATUS]">
-  <link href="../default.css" rel=stylesheet type="text/css">
-  <link href="../csslogo.ico" rel="shortcut icon" type="image/x-icon">
-  <link href="styles/styles-html.css" rel=stylesheet type="text/css">
-  <link href="[W3C-STYLESHEET-URL]" rel=stylesheet type="text/css">
+  <title>[TITLE] [LEVEL]</title>
```
tabatkins: If you want a different `<title>` and `<h1>`, please use the `Title` and `H1` metadata accordingly.  (I know, this was present previously, but might as well fix it now.)

bikeshed/include/html/header.include
```diff
@@ -2,13 +2,11 @@
...
+  <link href="https://www.w3.org/StyleSheets/TR/W3C-[STATUS]" rel=stylesheet type="text/css">
```
tabatkins: `[W3C-STYLESHEET-URL]` is correct - why did you change it?

bikeshed/include/html/header.include
```diff
@@ -2,13 +2,11 @@
...
+  <style data-fill-with="stylesheet">
```
tabatkins: This is clearly broken - no end tag.

bikeshed/include/html/copyright.include
tabatkins: Sorry; I meant specifically revert the loss of the trailing newline.  The rest of the copyright change is fine obviously. ^_^

bikeshed/include/html/header.include
arronei: The problem is the spec name is HTML the Level is 5.1. We want to define those as separate entities for use in other places in the document. However the `<title>` and the `<h1>` in the include file need to use the combined of what `<h1>`, and `Level:` defines in the `.bs` file. I would prefer to leave this alone for now since I am not doing major edits to the spec.

bikeshed/include/html/copyright.include
arronei: Weird I added the newline but github didn't catch the change here.

bikeshed/include/html/header.include
arronei: @tabatkins ok I fixed the boilerplate in HTML to use Title: HTML and removed the <H1> but I still need to define the &lt;title> and the &lt;h1> in the header file to be [TITLE] [LEVEL].

tabatkins: I'm still *super* confused.  What, precisely, are you trying to do?  The `Title` and `H1` metadata's *sole purpose* is to serve as appropriate things to fill the `<title>` and `<h1>` elements.  Why are you trying to set them to something different and then correct it in the template?

arronei: The title metadata for the document is "HTML" and the level is "5.1", However the &lt;title> and &lt;> is the combined title metadata and the level metadata because each of these metadata items is not completely accurate for the &lt;title> or the &lt;h1>. Additional places in the document use the [TITLE] and [LEVEL] metatdata items separately and not necessarily placed together. Making the title metadata "HTML 5.1" would be incorrect in those cases and all the cases since the title is really "HTML".

PR #668, opened 2016-04-15
<caption> must be the first child of <table>

mikewest:
Hooray for HTML validation. This blocks submission of specs through Specberus/Echidna.


mikewest:
I think the travis-ci test is broken on ToT without this patch. :/

tabatkins:
k, merged, and i'll see if things still need to be fixed up in the tests

PR #675, opened 2016-04-19
Remove test cases for exception types

arronei:
Correct test cases after change https://github.com/tabatkins/bikeshed/commit/ce3415237cd6224464d15e22b47d7fd2d909c993 where the exception type has been removed.


tabatkins:
Actually, removing the exception type was a mistake - I mistakenly thought exception types were interface-like, when they're actually a legacy enum-like thing.

I'll close this and re-add those exception types instead. ^_^

PR #682, opened 2016-04-22
Describe the TR link as the "latest published version".

jyasskin:
Since usually the ED will be more "latest" than the TR. This matches
ReSpec's wording.


tabatkins:
I'd have thought this would rub afoul of pubrules, but if ReSpec did it, I guess we're ok.

jyasskin:
I haven't checked pubrules, but https://www.w3.org/TR/permissions/ uses exactly this wording.

jeisinger:
hum, so 'specberus' doesn't like this :-/

jeisinger:
actually, disregard that, specberus seems to be fine

Issue #634, opened 2016-03-18
Bikeshed seems inconsistent with when method arguments need to be mentioned.


mkruisselbrink:
In the FileAPI bikeshed conversion there is IDL like this:
```idl
interface Blob {};
interface FileReader {
  void readAsText(Blob blob, optional DOMString label);
};
```

Furthermore there are reference to the method like `{{FileReader/readAsText()}}` and references to the methods arguments like `{{FileReader/readAsText()/blob}}`. The method references link correctly (despite absence of non-optional arguments), but the reference to the arguments complain that no definition is found unless I change them to `{{FileReader/readAsText(blob)/blob}}`. It seems to me like bikeshed should accept both or neither, not the current behavior where sometimes it's okay to leave out the arguments while other times it is not.

mkruisselbrink:
Actually ``{{FileReader/readAsText(blob)/blob}}` worked in my simplified sample, but in the actual spec only ``{{FileReader/readAsText(blob, label)/blob}}` works, presumably because of the explicit argument definition that's also in the spec:
```
<dfn argument for="FileReader/readAsText(blob, label)">blob</dfn>
```

Also unless I'm going crazy, yesterday this same spec compiled without errors, so something seems to have changed recently?

tabatkins:
You are correct, it's inconsistent and not a good situation.  This is because autolinking is ~magic~ and that magic is unreliable in some cases and hard to reason about, even for me.

I'm in the middle of refactoring autolinking right now to make this work a lot better.  I expect progress to be slow and gradual, but to complete the work in the next few weeks.  Apologies, and please bear with me while I make this all act more rational.

mkruisselbrink:
Actually my reduced sample still doesn't seem to match the spec. In my reduced sample (with above IDL and dfn tag) I can link to these method arguments, in the actual spec I don't seem to be able to link to them in any way I can come up with. `{{FileReader/readAsText(blob)/blob}}` fails with `LINK ERROR: No 'idl' refs found for 'blob' with for='FileReader/readAsText(blob)'.`, while `{{FileReader/readAsText(blob, label)/blob}}` fails with `LINK ERROR: No 'idl' refs found for 'readAsText(blob, label)/blob'.` (the "for='FileReader'" seems to have gotten lost somehow?)

birtles:
[...]

PR #684, opened 2016-04-28
Possible fix for some autolinking issues

birtles:
As discussed in issue #634


tabatkins:
I'm not 100% sure why I thought having a space in those patterns would *ever* work. I'll merge and see if anything bad happens.

PR #693, opened 2016-05-05
Fix Index Out of Range error for enums with "" empty string as values

### LINKED ISSUE #692: IndexError when trying to process enum with ""
```
Traceback (most recent call last):
  File "C:\Python27\Scripts\bikeshed-script.py", line 9, in <module>
    load_entry_point('Bikeshed==0.0.0', 'console_scripts', 'bikeshed')()
  File "d:\github\bikeshed-arronei\bikeshed\__init__.py", line 171, in main
    doc.preprocess()
  File "d:\github\bikeshed-arronei\bikeshed\__init__.py", line 502, in preprocess
    processIDL(self)
  File "d:\github\bikeshed-arronei\bikeshed\__init__.py", line 1716, in processIDL
    error=False)
  File "d:\github\bikeshed-arronei\bikeshed\ReferenceManager.py", line 221, in getRef
    variedTexts = [v for v in linkTextVariations(text, linkType) if v in self.defaultSpecs]
  File "d:\github\bikeshed-arronei\bikeshed\ReferenceManager.py", line 631, in linkTextVariations
    if str[0] != "_":
IndexError: string index out of range
```
I think we need a special case for this because how would I link to the value dfn for ""?

```html
<pre class='metadata'>
Title: Test
Status: WD
TR: http://www.w3.org/TR/SpecName/
ED: http://www.w3.org/TR/SpecName
Previous Version: http://www.w3.org/TR/SpecName
Shortname: SpecName
Level: 1
Editor: John Doe, AnOrg., AnOrg@mail.com
Abstract: Abstract sentence
</pre>

<pre class="idl" data-highlight="webidl" dfn-for="CanPlayTypeResult">
enum CanPlayTypeResult { "" /* empty string */, "maybe", "probably" };
</pre>
```

arronei:
This is a work around fix for the IndexError. This issue is blocking us from updating to the latest bikeshed. We have to use a snapshot from last month.

Fix #692

### DIFF
--- a/bikeshed/__init__.py
+++ b/bikeshed/__init__.py
@@ -1710,10 +1710,10 @@ def processIDL(doc):
                         forceDfn = True
                 else:
                     for linkFor in config.splitForValues(el.get('data-idl-for', '')) or [None]:
-                        ref = doc.refs.getRef(idlType, idlText,
-                                              linkFor=linkFor,
-                                              el=el,
-                                              error=False)
+                        try:
+                            ref = doc.refs.getRef(idlType, idlText, linkFor=linkFor, el=el, error=False)
+                        except IndexError:
+                            break
                         if ref:
                             url = ref.url
                             break


tabatkins:
This is a big hack, when the solution is to just do what I'm purposely doing for all the other linkTextVariation cases, and use a range rather than asking for a specific index.  Committing that now; closing this PR.

PR #696, opened 2016-05-06
Allow explicitly unleveled documents.

mikewest:
Currently, Bikeshed includes a level along with the shortname when generating
links for certain classes of documents. Based on the conversation at
https://lists.w3.org/Archives/Public/spec-prod/2016AprJun/0005.html, there are
at least some groups producing documents without levels which Bikeshed believes
ought to have levels. This patch allows those documents to explicitly note
that they are unleveled.

### DIFF
--- a/bikeshed/MetadataManager.py
+++ b/bikeshed/MetadataManager.py
@@ -16,7 +16,7 @@ from .repository import *
 class MetadataManager:
     @property
     def vshortname(self):
-        if self.level is not None:
+        if self.level is not None and len(self.level.nums):
             return "{0}-{1}".format(self.shortname, self.level)
         return self.shortname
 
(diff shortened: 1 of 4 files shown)


mikewest:
WDYT, @tabatkins?

tabatkins:
So basically you can put in whatever you want in the level field, as long as it's not numeric, and it'll consider you "unlevelled"?  I think I'd prefer blessing a term (like "none") so we don't squat on too much syntax space in the future.

tabatkins:
Feel free to make HierarchicalNumber handle this; you can add a __nonzero__() method to it to just make it falsey in that case, and then simplify the check in vshortname to just check for truthiness.

tabatkins:
Wrote it myself, closing this.

PR #700, opened 2016-05-11
Fix FXTF status include files

birtles:
This updates the FXTF status include files to appease specberus.

Specifically this regex:
https://github.com/w3c/specberus/blob/f16f7ce5f7970934cadcc13f8d87082c7e34ac95/lib/rules/sotd/pp.js#L49

This involves dropping the "(part of the ...)" parenthicals, as well as replacing "produced by" with "published by".

At the same time this updates the URLs in this paragraph to use https.


birtles:
Oops, don't know where those merge commits came from.

birtles:
Merge commits should be gone now (or I _completely_ messed things up).

tabatkins:
I always squash-merge now anyway, so it no longer matters. ^_^

PR #729, opened 2016-06-21
WebAppSec: Presubmit bits and pieces for Echidna.

mikewest:
WDYT, @tabatkins? This seems to work for me locally (https://www.w3.org/TR/2016/WD-CSP3-20160621/ went through, anyway...).


tabatkins:
If it works for you, it works for me.

Issue #736, opened 2016-07-13
C++ paper statuses


jfbastien:
As a continuation to #713, it would be useful to support C++ paper statuses.

i.e. document metadata:
```
<pre class='metadata'>
Title: Foo
Shortname: Bar
Status: ----> THIS <----
Group: WG21
URL: http://wg21.link/baz
Editor: JF Bastien, Google, jfb@google.com
Abstract: Hai
Date: 2016-07-13
Markup Shorthands: markdown yes
</pre>
```

Right now this seems hardcoded in bikeshed across all groups. It may need to move to boilerplate?

Consulting the [isocpp docs](https://isocpp.org/std/standing-documents/sd-7-mailing-procedures-and-how-to-write-papers) I think the statuses that make sense are:

* Issue
* Defect Report
* Draft (D numbered paper)
* Published (P numbered paper)
* Meeting announcements
* Minutes
* Working Drafts
* Editor’s Report
* CD
* PDTS
* DTS
* Records of Response

Tagging @jyasskin and @brycelelbach for input.

tabatkins:
Cool. So all I need, I guess, is a short name for each (to use in the `Status` metadata), and what you want to show up in the subheading for each.

brycelelbach:
I - Issue
DR - Defect Report
D - Draft Proposal 
P - Published Proposal 
MEET - Meeting Announcements
RESP - Records of Response
MIN - Minutes
ER - Editor's Report
SD - Standing Document
PWI - Preliminary Work Item
NP - New Proposal
NWIP - New Work Item Proposal
WD - Working Draft
CD - Committee Draft
FCD - Final Committee Draft
DIS - Draft International Standard
FDIS - Final Draft International Standard
PRF - Proof of a new International Standard
IS - International Standard
TR - Technical Report
DTR - Draft Technical Report
TS - Technical Specification
DTS - Draft Technical Specification
PAS - Publicly Available Specification
TTA - Technology Trends Assessment
IWA - International Workshop Agreement
COR - Technical Corrigendum
GUIDE - Guidance to Technical Committees
NP-AMD - New Proposal Amendment
AWI-AMD - Approved new Work Item Amendment
WD-AMD - Working Draft Amendment
CD-AMD - Committe Draft Amendment
PD-AMD - Proposed Draft Amendment
FPD-AMD - Final Proposed Draft Amendment
D-AMD - Draft Amendment
FD-AMD - Final Draft Amendment
[...]

PR #745, opened 2016-07-14
Drop the space in iso/D iso/P

jfbastien:
It's cleaner.

Follow-up to #736.


tabatkins:
Mind fixing iso/I too?

jfbastien:
> Mind fixing iso/I too?

Done.

PR #754, opened 2016-07-15
Add 'Hide deleted text' checkbox

### LINKED ISSUE #744: Spec diff support
We love `<ins>` and `<del>`, but sometimes you don't want to see `<del>` and you just want `<ins>`.

SPEC DIFF TO THE RESCUE!

With SPEC DIFF you can just show the spec as it would look!
e.g. that little checkbox at the top of http://wg21.link/n4430

(SPEC DIFF is a horrible misnomer)
(SPEC DIFF may lead to drowsiness)
(as your doctor if SPEC DIFF is right for you)

jfbastien:
Resolves #744.


jfbastien:
The CSS selector isn't not quite right, and @tabatkins had mentioned that he wanted a boilerplate thing instead of a new type of metadata. This kinda does the right thing otherwise.

bikeshed/MetadataManager.py
```diff
@@ -744,6 +745,7 @@ def joinDdList(a,b):
     "Repository": Metadata("Repository", "repository", joinValue, parseRepository),
     "Revision": Metadata("Revision", "level", joinValue, parseLevel),
     "Shortname": Metadata("Shortname", "shortname", joinValue, parseLiteral),
+    "Show Hide Deleted Button": Metadata("Show 'Hide deleted text' button", "showHideDeletedButton", joinValue, parseBoolean),
```
tabatkins: The first arg is what you actually write in your metadata; is this what you really want people to write?

jfbastien: That's what I have right now in my paper. No particular love for it versus anything else. Suggestion? "Hide deleted text"?

tabatkins: That's not accurate. I'm thinking: `Toggle Diffs`?  Oh, and the key and first argument have to be identical; it's unfortunately repeated information, but I can't remove the duplication cleanly so eh.

jfbastien: Yeah that's better, renamed.

jfbastien:
Ping!

bikeshed/MetadataManager.py
```diff
@@ -70,6 +70,7 @@ def __init__(self, doc):
         self.statusText = []
         self.testSuite = None
         self.title = None
+        self.toggleDiffs = None
```
tabatkins: False here, since this is a bool and false is the default value.

bikeshed/boilerplate.py
```diff
@@ -715,4 +726,3 @@ def addIssuesSection(doc):
         del idel.attrib['id']
     for dfnel in findAll(config.dfnElementsSelector, container):
         dfnel.tag = "span"
-
```
tabatkins: Lost the final linebreak here. I'll just put it back next time I touch this file, so let's avoid diff churn. ^_^

tabatkins:
Ready to merge after fixing nits.

jfbastien:
Done. Double line at EOF is weird!

tabatkins:
It's not a double line, it just ensures that every line ends in a `\n`.  It's standard practice!

PR #760, opened 2016-07-20
Markup 'Assert:' as an assertion box.

mikewest:
@tabatkins: WDYT? I've started using `Assert: This is an assertion!` in algorithms, and I've seen them pop up in other specs (like HTML) as well; it would be nice to highlight them somehow. Here I've just copied the advisement styling, but perhaps you CSS folks have a sense of style that might guide you to something prettier?

You can see the way it looks in https://w3c.github.io/webappsec-secure-contexts/#settings-object.

### DIFF
--- a/bikeshed/MetadataManager.py
+++ b/bikeshed/MetadataManager.py
@@ -35,6 +35,7 @@ class MetadataManager:
         self.rawStatus = None
 
         # optional metadata
+        self.assertionClass = "assertion"
         self.advisementClass = "advisement"
         self.atRisk = []
         self.audience = []
@@ -666,6 +667,9 @@ def parseDoc(doc):
     for el in findAll(".replace-with-issue-class", doc):
         removeClass(el, "replace-with-issue-class")
         addClass(el, doc.md.issueClass)
+    for el in findAll(".replace-with-assertion-class", doc):
+        removeClass(el, "replace-with-assertion-class")
+        addClass(el, doc.md.assertionClass)
     for el in findAll(".replace-with-advisement-class", doc):
         removeClass(el, "replace-with-advisement-class")
         addClass(el, doc.md.advisementClass)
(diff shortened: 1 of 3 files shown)


domenic:
-1 from me on the look, I don't think assertions deserve such strong callouts (using the same visual style as warnings). To me they're like extra-boring non-normative notes.

But it seems reasonable to give them an "assertion" class so people can experiment with different styles, as long as the styles are not forced on every group. (I can't tell whether stylesheet.include is for every group or is just the default/W3C stylesheet...)

mikewest:
`stylesheet.include` is on for every group. I'm totally happy to move the styles out of the global stylesheet, because I think I disagree about their importance. Like, if I could make `DCHECK` in Chrome's code bright red, I totally would. ;)

bikeshed/MetadataManager.py
```diff
@@ -708,6 +712,7 @@ def joinDdList(a,b):
 
 knownKeys = {
     "Abstract": Metadata("Abstract", "abstract", joinList, parseLiteralList),
+    "Assertion Class": Metadata("Assertion Class", "assertionClass", joinValue, parseLiteral),
```
tabatkins: Alphabetize, please?

bikeshed/markdown.py
```diff
@@ -337,6 +337,8 @@ def parseParagraph(stream):
 	elif line.lower().startswith("issue:"):
 		line = line[6:]
 		p = "<p class='replace-with-issue-class'>"
+        elif line.lower().startswith("assert:"):
```
tabatkins: indent looks to be wrong here

tabatkins:
Well, `stylesheet.include` is just one of the boilerplate files, which is used by the default boilerplate and many of the group boilerplates.

I gave some nits; with those fixed, I'm happy to merge.  I don't have a strong opinion on the styles, but people can always override if they want (or just swap the assertion class to something unique and style it without conflict).

bikeshed/markdown.py
mikewest: You use tabs because of your name, don't you? :)

bikeshed/MetadataManager.py
mikewest: Done.

mikewest:
Softened the style to something grey as opposed to orange, and fixed up the nits. WDYT?

bikeshed/markdown.py
tabatkins: Huh. Actually, no, I write my Python with spaces per PEP8, but this file must have slipped tabs in without me realizing. I'll make sure everything is consistent in a bit. ^_^

PR #762, opened 2016-07-21
Use the entire translated ID rather than just the first character of it.

### LINKED ISSUE #761: Translate Ids metadata stopped working correctly
With for example `Translate Ids: foobar dfn-foobar` bikeshed ends up translating `foobar` to `d` rather than to `dfn-foobar`. From a git bisect this started failing in any of the following commits (can't narrow it down further since at many of those commits bikeshed just crashes):
9c1ab493e9279662513783c4112249eb8a10ff30
33baa315b2c89df98517f442d724d05fb4e54b0e
fb5288e938c3889ca93cbc1ba871b62c84a5b9f7
6790ed8301ea3610bc2e8e9af41c8fa71e6a271f
7dd9ac64d7622e9a2ffe26d105b306b6af20b6fa
c782f396e0cc3eed5fa3e478f343ee2df5f452ba

mkruisselbrink:
This fixes #761

### DIFF
--- a/bikeshed/htmlhelpers.py
+++ b/bikeshed/htmlhelpers.py
@@ -459,7 +459,7 @@ def fixupIDs(doc, els):
 def translateIDs(trans, els):
     for el in els:
         if el.get('id') in trans:
-            el.set('id', trans[el.get('id')][0])
+            el.set('id', trans[el.get('id')])
 
 def dedupIDs(doc):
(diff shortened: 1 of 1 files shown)


tabatkins:
Ahaha, I must have had it as a tuple earlier, and then changed to a bare string and didn't update.

PR #764, opened 2016-07-25
htmlserializer has moved

philandstuff:
html5lib/html5lib-python@6c30d0bd7d5dfa4722368a8dabc637d057350185
(first released in html5lib 0.99999999) changed the location of the
serializer namespace, causing bikeshed to break.  This updates bikeshed
to use the new location of the namespace.

### DIFF
--- a/bikeshed/htmlhelpers.py
+++ b/bikeshed/htmlhelpers.py
@@ -3,7 +3,7 @@ from __future__ import division, unicode_literals
 import hashlib
 import html5lib
 from html5lib import treewalkers
-from html5lib.serializer import htmlserializer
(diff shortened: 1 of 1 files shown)


tabatkins:
Hmm, I'll need to do some conditional importing instead, because I know people are still using the old html5lib and there's no pressing need for them to update.

Or I can just excise this last remaining use of the built-in serializer, I suppose.

philandstuff:
yup, makes sense. 👍

tabatkins:
Ah, nm, I'm not even using that import anymore! Just deleted the line instead. ^_^

PR #780, opened 2016-08-09
Fix flake8 W391 blank line at end of file

jfbastien:


tabatkins:
That one we'll keep on the ignore list - it's intentional (and my text editor is set up to add it to all files).

tabatkins:
nm, didn't see that it was *two* empty lines there.

PR #781, opened 2016-08-09
Fix flake8 W191 indentation contains tabs

jfbastien:


jfbastien:
Rebased, should be good to go now.

PR #783, opened 2016-08-10
More flake8

jfbastien:
These are *all* cosmetic changes where PEP8 is like "whitespace is so insignificant that I want it to look exactly my way". I didn't fix all of them (ignored the ones with too many hits, stopped after a few to avoid huge spam), all of the fixes are automated, and I split each into its own patch (so you can audit each, and merge all as one or separately as you wish).


tabatkins:
I'll probably continue to mess up the "2 spaces before inline comment" rule, but whatever. The rest look fine to me.

PR #784, opened 2016-08-11
Yet more flake8 fixes

jfbastien:
This contains some of the wordier whitespace fixes from flake8.

The remainder are the ones I think you want to keep as-is:
* E203 whitespace before ':'
* E231 missing whitespace after ':'
* E501 line too long (89 > 79 characters)
* F403 'from messages import *' used; unable to detect undefined names


tabatkins:
Yup, these all look good, and you're correct that I want to keep those exceptions.

PR #785, opened 2016-08-11
Inline markdown elements: emphasis, strong, link

### LINKED ISSUE #778: More markdown support?
I'm trying to understand why bikeshed supports [a subset of markdown](https://github.com/tabatkins/bikeshed/blob/master/docs/markup.md) and not other parts. I'm not sure I get a trend beyond what's currently implemented is what's there.

 Specifically, I'd like to support:
* Regular markdown links à la `[awesome](zombo.com "Anything is possible")` as well as with ID.
* Emphasis and bold with `*` and `**`.

Is that just something I can send a PR for, or do you have a moral objection?

jfbastien:
Fixes #778.

### DIFF
--- a/bikeshed/shorthands.py
+++ b/bikeshed/shorthands.py
@@ -226,6 +226,16 @@ def transformAutolinkShortcuts(doc):
     def varReplacer(match):
         return E.var(match.group(1))
 
+    strongRe = re.compile(r"([_\*])\1([\d\w\s]+)\1\1")
+
+    def strongReplacer(match):
+        return E.strong(match.group(2))
+
+    emRe = re.compile(r"([_\*])([\d\w\s]+)\1")
+
+    def emReplacer(match):
+        return E.em(match.group(2))
+
     def transformElement(parentEl):
         processContents = isElement(parentEl) and not doc.isOpaqueElement(parentEl)
         if not processContents:
@@ -255,6 +265,9 @@ def transformAutolinkShortcuts(doc):
             config.processTextNodes(nodes, sectionRe, sectionReplacer)
         if "algorithm" in doc.md.markupShorthands:
             config.processTextNodes(nodes, varRe, varReplacer)
+        if "markdown" in doc.md.markupShorthands:
+            config.processTextNodes(nodes, strongRe, strongReplacer)
+            config.processTextNodes(nodes, emRe, emReplacer)
         return nodes
 
     transformElement(doc.document.getroot())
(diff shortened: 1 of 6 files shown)


tests/markdown011.html
```diff
@@ -0,0 +1,186 @@
...
+  <div data-fill-with="at-risk"></div>
+  <nav data-fill-with="table-of-contents" id="toc">
+   <h2 class="no-num no-toc no-ref" id="contents">Table of Contents</h2>
+  </nav>
+  <main>
+   <p>Text is <em>wonderful</em>. <strong>Amazing</strong> even.</p>
+   <p>Text is <em>wonderful</em>. <strong>Amazing</strong> even.</p>
+   <p><strong>SO GREAT</strong></p>
+   <p>un<em>frigging</em>believable</p>
+   <p>You can \*escape\* things. Except <strong>death</strong> and <em>taxes</em>.</p>
```
mkruisselbrink: It isn't really escaping things if that leaves the escape character in the resulting output...

bikeshed/shorthands.py
```diff
@@ -226,6 +226,16 @@ def elementReplacer(match):
     def varReplacer(match):
         return E.var(match.group(1))
 
+    strongRe = re.compile(r"([_\*])\1([\d\w\s]+)\1\1")
```
tabatkins: No need to escape a `*` inside of a character class.

`\d\w\s` does not capture the full range of things you can put in there - for example, it excludes hyphens. You're gonna want a lookahead that just checks that the next two characters aren't `\1\1`.

bikeshed/shorthands.py
```diff
@@ -226,6 +226,16 @@ def elementReplacer(match):
...
+
+    def strongReplacer(match):
+        return E.strong(match.group(2))
+
+    emRe = re.compile(r"([_\*])([\d\w\s]+)\1")
```
tabatkins: Same as above.

jfbastien:
I addressed a bunch of the comments, and followed some of commonmark:
http://spec.commonmark.org/0.26/#emphasis-and-strong-emphasis

I ignored the punctuation character part of delimiter runs, and can't handle nesting because that's a limitation in the current bikeshed code.

tests/markdown011.html
jfbastien: Done.

bikeshed/shorthands.py
jfbastien: Done.

bikeshed/shorthands.py
jfbastien: Done.

jfbastien:
I added support for inline links as well. Again, not full markdown but close.

PR #825, opened 2016-09-29
Docker container

kba:
Here's an extension of @DavidBruant's earlier Docker PR #817.

This creates a fully self-contained docker container image that can be used just like the bikeshed command without the need to clone the repository or to install any python libraries.

I've set my fork up [automatically build on Docker Hub](https://hub.docker.com/r/kbai/bikeshed/tags/) so it can be run with

```sh
docker run --rm -it -v $PWD:/data kbai/bikeshed Overview.bs
```

from a folder containing `Overview.bs`.

Should you decide to merge this, you can set up a docker hub account and create an automated build that rebuilds the image for every push and run `sed -i s/kbai/your-dockerhub-account-name/ *`.


kba:
We're using this with this Makefile:

```make
BIKESHED = @$(shell \
		   if type bikeshed >/dev/null 2>&1;then \
				echo 'bikeshed';\
			elif type docker >/dev/null 2>&1;then \
				echo 'docker run --rm -it -v $(PWD):/data kbai/bikeshed'; \
			else \
				echo 'echo "bikeshed and docker not installed";exit 1;';\
			fi)
index.html: index.bs
	$(BIKESHED) -f spec index.bs
```
This will first look for bikeshed available locally, then fall back to the docker version.

DavidBruant:
> I've set my fork up automatically build on Docker Hub 

My understanding of how this works is that the image is updated when there are new commits in the main branch. However, this means the image won't be updated when the data pulled during a `bikeshed --update` are updated?

kba:
Right, I don't know when these change, it's not handled. But if there is a mechanism to be notified when that happens, you can trigger an automated build via Web hook (i.e.g a HTTP POST to docker hub).

PR #828, opened 2016-10-01
addCustomBoilerplate had reference to undefined var 'target'

kba:
https://github.com/tabatkins/bikeshed/commit/bdec26aaef62eaccaab681c49f76145df6700641 breaks the `addCustomBoilerplate` method since `target` is now undefined.

Not sure whether this is the right behavior now but without working `addCustomBoilerplate`, the `boilerplate-substitution001.bs` test will break.

### DIFF
--- a/bikeshed/boilerplate.py
+++ b/bikeshed/boilerplate.py
@@ -107,7 +107,8 @@ def addStyles(doc):
 def addCustomBoilerplate(doc):
     for el in findAll('[boilerplate]', doc):
         bType = el.get('boilerplate')
-        if bType in doc.fillContainers:
+        if bType in doc.fillContainers and len(doc.fillContainers[bType]) > 0:
+            target = doc.fillContainers[bType][0]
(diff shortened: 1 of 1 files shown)


tabatkins:
Whoops, indeed! The correct fix is to add

```udiff
         if bType in doc.fillContainers:
+            target = doc.fillContainers[bType][0]
             replaceContents(target, el)
             removeNode(el)
```

kba:
Thanks, updated the commit.

kba:
No, still not precise enough, `doc.fillContainers[bType]` can be an empty array. https://travis-ci.org/tabatkins/bikeshed/builds/164283149

tabatkins:
Ugh, you're right. I forgot that defaultdicts can be "poisoned" by just looking for a key, so you have to assume that they always have a value.  I've gone ahead and fixed things appropriately in bdec26a, so I'll close this PR.

PR #843, opened 2016-10-21
Option to inline github issue titles only

kba:
With `Inline Github Issue Titles: on`, for

> Issue(2):

will output

> Issue 8  tabatkins/bikeshed#2  Add a setup.py

instead of

> ISSUE 2 ON GITHUB "ADD A SETUP.PY"
> 
> This setuptools/pip will automatically install the dependencies and an executable script.
> 
> With this file in the repository, pip install
> git+https://github.com/tabatkins/css-preprocessor.git should be enough
> to install the preprocessor on a system with pip, python-dev,
> libxml2-dev, and libxslt1-dev.


bikeshed/MetadataManager.py
```diff
@@ -816,6 +817,7 @@ def parseLiteralList(k, v, l):
     "Indent": Metadata("Indent", "indent", joinValue, parseInteger),
     "Infer Css Dfns": Metadata("Infer Css Dfns", "inferCSSDfns", joinValue, parseBoolean),
     "Inline Github Issues": Metadata("Inline Github Issues", "inlineGithubIssues", joinValue, parseBoolean),
+    "Inline Github Issue Titles": Metadata("Inline Github Issue Titles", "inlineGithubIssueTitles", joinValue, parseBoolean),
```
kba: Good idea, will do.

PR #896, opened 2017-01-12
Provide better error messages for ambiguous for-less links

dbaron:
This sequence of four commits (most of which refactor things) provides more verbose error messages for ambiguous for-less links (but also doesn't give the element twice).  It uses the reference syntax to report what the ambiguous references are.

I concede that my new function names may not be that great, nor perhaps is the way of passing a bunch of parameters to some of these.  Let me know if there's something you want me to fix about those.

I don't see any tests of error reporting in bikeshed, so it's not clear to me how to add a test for this.  Maybe I'm missing something, though.

### DIFF
--- a/bikeshed/ReferenceManager.py
+++ b/bikeshed/ReferenceManager.py
@@ -452,13 +452,7 @@ class ReferenceManager(object):
         if len(blockRefs) == 1:
             return blockRefs[0]
         elif len(blockRefs) > 1:
-            possibleRefs = [refToText(ref) for ref in simplifyPossibleRefs(blockRefs)]
-            linkerror("Multiple possible '{0}' refs for '{1}'.\nArbitrarily chose the one in {2}.\nIf this is wrong, insert one of the following lines into a <pre class=link-defaults> block:\n{3}",
-                      linkType,
-                      text,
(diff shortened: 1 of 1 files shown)


dbaron:
As an example, this patch changes the [error that I just fixed](https://hg.csswg.org/drafts/rev/da7224e63bb2) from reporting the message:

```
LINK ERROR: Ambiguous for-less link, please see <https://tabatkins.github.io/bikeshed/#ambi-for> for instructions:
<a class="css" data-link-type="maybe" data-lt="dpi">dpi</a>
<a class="css" data-link-type="maybe" data-lt="dpi">dpi</a>
```

to reporting the message:

```
LINK ERROR: Ambiguous for-less link for 'dpi', please see <https://tabatkins.github.io/bikeshed/#ambi-for> for instructions:
Local references:
spec:css-values-4; type:value; for:<resolution>; text:dpi
for-less references:
spec:css-images-3; type:value; for:/; text:dpi
<a class="css" data-link-type="maybe" data-lt="dpi">dpi</a>
```

dbaron:
Er, there's a bug in one of the refactoring commits that I just noticed -- I'll need to fix that up.

PR #897, opened 2017-01-13
Allow "Note" to be separately stylable

arronei:
- Add tag around "Note" text to make it separately styleable if needed
-Updated test case

### DIFF
--- a/bikeshed/markdown.py
+++ b/bikeshed/markdown.py
@@ -342,8 +342,12 @@ def parseParagraph(stream):
     line = stream.currtext()
     initialPrefixLen = stream.currprefixlen()
     endTag = "</p>"
-    if line.lower().startswith("note:") or line.lower().startswith("note, "):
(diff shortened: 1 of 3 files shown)


arronei:
We could simplify this and just remove the word r"Note[:,]]\s*" like all the other cases. It just seems weird that this was the only case where the identifying text wasn't removed. This is why this change works the way it does by matching and keeping the text as it was.

tabatkins:
I leave the "Note" text in so you can have a note starting with "Note, blah blah blah." - removing it and fixing up the sentence it was a part of is non-trivial.  That's not an issue with "Issue:", as that doesn't look like it's starting an English sentence; people naturally start a fresh sentence after it.

Anyway, patch looks fine to me. Merging.

domenic:
What would be nice is removing "Note:" or maybe just "NOTE:" (but leaving "Note " or "Note, " alone).

Issue #894, opened 2017-01-09
Better support for defining internal methods and slots


domenic:
Continued from https://github.com/whatwg/html/pull/2145; impacts https://github.com/whatwg/html/pull/2227 ; /cc @isonmad

Here is the test document:

```
<pre class="metadata">
Title: Test
Group: WHATWG
Shortname: test
Status: LS
No Editor: true
Abstract: blah
</pre>

Platform objects can have a \[[<dfn data-export="" data-dfn-type="method" data-dfn-for="platform objects">Transfer</dfn>]]() internal method.

Platform objects can have a \[[<dfn data-export="" data-dfn-type="attribute" data-dfn-for="platform objects">Detached</dfn>]] internal slot.

The \[[<a method for="platform objects">Transfer</a>]]() method is defined as follows:

1. Set the \[[<a attribute for="platform objects">Detached</a>]] internal slot to true.

That is, the \[[{{platform objects/Transfer}}]]() method sets the \[[{{platform objects/Detached}}]] internal slot.
```

Currently it fails saying

```
[1;31mFATAL ERROR:[0m Functions/methods must end with () in their linking text, got 'Transfer'.
```

Adding `data-lt="[[Transfer]]()"` does not help:

```
[1;31mFATAL ERROR:[0m Functions/methods must end with () in their linking text, got '[[Transfer]]()'.
```

Also of note, so far in HTML we've decided that the [[]]s are not part of the method/slot name, (e.g. they are not bold in the definition and not linked at the usage site).

I suspect even if we got this to work, the shorthand would be monospaced, which is again the wrong typographic convention.

Maybe this merits some first-class support after all? Of note, internal slots and internal methods are in the same namespace, but then again, so are IDL attributes and IDL methods.

isonmad:
Ping.

tabatkins:
> Also of note, so far in HTML we've decided that the [[]]s are not part of the method/slot name, (e.g. they are not bold in the definition and not linked at the usage site).

Hmm, that's exactly opposite the tradition I've already tuned Bikeshed to recognize.  If you do `<dfn attribute>\[[Foo]]</dfn>` it'll use a slightly different ID pattern, and `{{[[Foo]]}}` is explicitly handled and doesn't need escaping.  I'll need to tune things a little to make it recognize `[[Foo]]()` as an internal method, but if y'all stuck with that, it would make things a lot easier.

tabatkins:
[...]

PR #898, opened 2017-01-17
add internal-slot dfn type

isonmad:
For #894.

Not making a separate type for internal-method, because "attribute" and "method" are only separate for the sake of enforcing the latter ending in `()` (as it's in `functionishTypes`). It could be added later, but with the convention being to put things outside the linking text like `[[<a>InternalMethod</a>]]()` it wouldn't work as is.


tabatkins:
Since #894 was closed no change, I'll close this too.

PR #903, opened 2017-01-21
Add basic support for TC39

jfbastien:
Stages are documented here: https://tc39.github.io/process-document/
ECMArkup is a tool similar to bikeshed which handles more of TC39 things: https://bterlson.github.io/ecmarkup/


jfbastien:
Oh hmm things changed place since I last updated. Will need to fix that, but g2g for now.

jfbastien:
@tabatkins updated!

PR #907, opened 2017-01-24
Master

garykac:
Add support for "stylesheet-extra" fill container

This is useful when you want to use Bikeshed’s default stylesheet, but
also have some additional custom CSS.

Currently, if you create a “stylesheet.include” file for a spec, then it will override the entire Bikeshed stylesheet. :-(

This is step 1 to switching the UIEvents specs to be part of the web platform group (and then removing the “uievents” group from bikeshed).


domenic:
You can just put `<style>` or `<link>` elements in your Bikeshed document...

garykac:
Argh! I didn't even try that.

Apparently, I think of Bikeshed as more as a Markdown-like editor and forget that I can slap HTML wherever I like.

tabatkins:
^_^

(In the future, it's usually best to hit me with an Issue first, to check if I already have an idea on how to accomplish whatever you're thinking of.)

PR #911, opened 2017-01-30
Add echidna parser options

tobie:
* --self-contained indicates the tar file should not
  bundle any extra directories.
* --additional-directories bundles directories in the tar file.
  This is useful to include images, etc. which are referenced
  by the specification.
  Defaults to images/, diagrams/ and examples/.


tabatkins:
These two flags are already handled by the bikeshed-extensions.include file. Is there a particular reason you're trying to add them to the flags?

tobie:
Oh, I totally missed that. Found it now. Thanks!

tobie:
Reopening this, as I'm now realizing your solution is a per WG one. I really need this as a per spec option. I currently have a 2 documents in the same folder. One that must include stuff from the examples directory the other one that shouldn't. I'm not sure how to resolve this by any other means than CLI options.

tabatkins:
Makes sense. Merged.

PR #918, opened 2017-02-04
Add installation instructions for Termux

frivoal:


tabatkins:
Holy shit, didn't realize this was possible. I, uh, might be doing this on my tablet soon.

frivoal:
It is possible and great fun. The only small annoyance is that your home folder in termux is not readable by other applications on the phone/tablet, so you have to put your specs under `~/storage/shared/<whatever-you-like>` if you want to be able to reach them via `file://sdcard/<whatever-you-like>` in the browser. It's not a big deal, but that's putting some constraints on where you can store your stuff.

A `bikeshed serve` command that does like `bikeshed watch` **and** sets up an http server on 127.0.0.1:8000 would make irrelevant thought. wink wink nudge nudge.

Issue #923, opened 2017-02-08
Use LXML wheels for Windows in install guide


TBBle:
It's not clear to me why this didn't go ahead under #506, but as far as I can tell, [LXML 3.6.0](https://pypi.python.org/pypi/lxml/3.6.0) has working wheels for both 32-bit and 64-bit Python 2.7, which simplifies the installation to be very similar to other PIP-based platforms.

The following commands worked for me. I did it in a virtualenv environment but that shouldn't affect the results.
```cmd
; Create my Python 2 virtualenv, not needed but good practice.
py -2 -m virtualenv venv
venv\Scripts\activate
; The actual installation process per docs/install.md
pip install pygments lxml==3.6.0
git clone https://github.com/tabatkins/bikeshed.git
pip install --editable bikeshed
bikeshed update
```

I'll put together a pull request to update docs/install.md appropriately.

Newer versions of LXML have wheels for Python 3.5, but not Python 2.7. Since the current installation instructions point at LXML 3.4.4 on Windows and _latest_ (3.7.2 right now) on other platforms, I assume we're not tied heavily to LXML versions.

(To save any hunting for others, the Python 2.7 wheels don't build due to http://stackoverflow.com/a/41071796/166389 which appears to be a bug in Python 2.7.12)

gregwhitworth:
Primarily because I couldn't get them running and didn't have the time to mess with it when I had something that worked. Thanks for the docs - I look forward to using this over the exe installs.

PR #925, opened 2017-02-08
Add build script for AppVeyor cloud CI system

TBBle:
This runs the testsuite against both 32-bit and 64-bit Python 2.7 on Windows.

Note: [Badges](https://www.appveyor.com/docs/status-badges/) are available in AppVeyor, but the URL is only available to the project owner, so I couldn't add it as part of this push request.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,20 @@
+# See https://packaging.python.org/appveyor/
+
+environment:
+  matrix:
+  - PYTHON: C:\\Python27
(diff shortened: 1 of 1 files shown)


tabatkins:
Cool, I'll trust you that this works. 😀

TBBle:
You certainly may trust me. Or (if you consider it useful) you could hook up the project master branch to AppVeyor auto-building like we have for Travis. Then we get an extra status badge, and it also keeps the Windows install instructions (#923) honest.

Issue #922, opened 2017-02-07
Echidna has access issues on Windows and throws Error 32


gregwhitworth:
Just so that I can keep track of this, and maybe someone has a silver bullet for me, when running `bikeshed echidna` it is throwing the following error:

    WindowsError: [Error 32] The process cannot access the file because it is being used by another process: 'c:\\users\\[username]\\appdata\\local\\temp\\tmpwj2af6'

If I'm able to find a solution I'll provide it here and update the docs.

PR #926, opened 2017-02-08
Fixed access issues #922 with temp file on Windows

gregwhitworth:
Windows leave's the temp file open while creating the tar which is causing the issue. Closing it makes the error go away and the rest of the process to resume.


gregwhitworth:
This fixes issue: https://github.com/tabatkins/bikeshed/issues/922

PR #924, opened 2017-02-08
Install LXML 3.6.0 from PIP for Windows

TBBle:
This updates the instructions as described in tabatkins/bikeshed#923, bumps the link for 'Latest' Python to the current latest, and applies some minor cosmetic fixes I noticed while I was there.

PR #928, opened 2017-02-10
Remove docs/install.md

TBBle:
The contents have been migrated into docs/index.bs.

Mentioned in tabatkins/bikeshed#924


TBBle:
I went with a remote link (https://tabatkins.github.io/bikeshed/#installing) rather than a local link (docs/index.html#installing) because both of the existing references appear in files with other references to that document.

PR #939, opened 2017-02-28
Use Process 2017

plehegar:
Must be deployed between March 1 and March 31.


plehegar:
I'm guessing that the failure is false report otherwise I'm not sure why it's triggered...

tabatkins:
Yeah, I'm messing around with some code that's causing spurious test
failures at the moment, so don't trust Travis.

PR #955, opened 2017-03-15
Fix boilerplate for Media Capture Task Force specs to points to right groups

dontcallmedom:


tabatkins:
Why'd you totally change the formatting of the paragraph?

dontcallmedom:
the existing formatting wasn't great?

dontcallmedom:
and also, I copied & pasted from existing docs where the right links/wording existed

tabatkins:
The existing formatting was using semantic linebreaking, so you could actually tell what lines changed; now all I can see is that something in the paragraph is different. It's also now inconsistent markup formatting with the rest of the file.

I'll merge, this sort of thing just makes me sad. 😔

dontcallmedom:
ha, sorry, I guess semantic linebreaking was a pretty good thing to have - I wasn't aware this was the adopted practice for bikeshed.

I'm happy to provide another PR that brings it back.

PR #961, opened 2017-03-22
Allow to configure a canonical URL for bikeshed documents

### LINKED ISSUE #959: Allow to define the rel=canonical URL
[ReSpec allows to set a canonical URL](https://github.com/w3c/respec/pull/938) for specs, and as [discussed in spec-prod](https://lists.w3.org/Archives/Public/spec-prod/2017JanMar/0008.html), I would like to bring the same feature to Bikeshed.

Since that'd would be my first code contribution to bikeshed, I would like to get some guidance on how to it right; my guess is that I need to update `MetadataManager.py` to detect the right configuration key, and then `boilerplate.py` to insert the `<link rel=canonical>` markup in the `<head>`.

Is that about right? any other thing I should pay special attention to?

dontcallmedom:
close #959

### DIFF
--- a/bikeshed/MetadataManager.py
+++ b/bikeshed/MetadataManager.py
@@ -43,6 +43,7 @@ class MetadataManager:
         self.audience = []
         self.blockElements = []
         self.boilerplate = config.BoolSet(default=True)
+        self.canonicalURL = None
         self.canIUseURLs = []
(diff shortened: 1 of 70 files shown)


dontcallmedom:
I'm not sure why the `include*` tests fail.

dontcallmedom:
tests now passing

PR #1007, opened 2017-04-27
Mention support for monospaced Markdown span elements

TimothyGu:


tabatkins:
The significance of `` `foo` `` is more than just monospace. I ended up rewriting that line a little bit anyway, and put the inline-code shortcut as a separate bullet point.

TimothyGu:
Okay, thanks!

PR #1009, opened 2017-05-05
Update copyright to point to correct license

svgeesus:
Per 2016 CSS WG charter, we publish under the more permissive "Document and Software license" not the "Document License".
https://www.w3.org/Style/2016/css-2016.html#licensing
Thus, changed to
https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document
instead of
https://www.w3.org/Consortium/Legal/2015/doc-license

Also updated w3c urls to https.


svgeesus:
The failing checks seem unrelated to what I changed :)

tabatkins:
Right, I've got two tests that use the CSSWG group, for more intensive integration testing, and this change'll definitely make them fail. No problem, but I won't be able to merge this until Monday when I'm back in front of a real computer and can update test expectations.

tabatkins:
Sorry for the delay, but this is merged now, along with a test rebase.

PR #1011, opened 2017-05-08
Change rel=previous to rel=prev

xfq:
According to the HTML Standard, `rel=previous` is "to be handled as specified by user agents, but must not be used in documents".

I'm not quite familiar with Python, so if I made any mistake, please point it out. Thanks!

(validator/validator#254 has some background information.)


tabatkins:
Whoops, not sure how I missed this PR. I merged the other one instead, so I'll close this one. Thanks!

xfq:
> Whoops, not sure how I missed this PR. I merged the other one instead, so I'll close this one. Thanks!

You're welcome. As Florian [said](https://github.com/w3c/csswg-drafts/pull/1177#issuecomment-293115803), "the effect is the same" :)

PR #1015, opened 2017-05-10
Use rel=prev instead of rel=previous

zcorpan:
https://checker.html5.org/ gives this warning:

> Warning: Potentially bad value previous for attribute rel on element a: The keyword previous for the rel attribute should not be used. Consider using prev instead.


xfq:
See #1011 :-)

PR #1016, opened 2017-05-12
Use HTTPS for Can I use... URLs

xfq:
See https://caniuse.com/feed/158


xfq:
Maybe I should also update the test `.bs` file...

PR #1018, opened 2017-05-13
Remove sudo from Travis testing script

TBBle:
Travis runs Python in a virtualenv, so there's no need for sudo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,10 @@ language: python
 python:
  - "2.7"
 before_install:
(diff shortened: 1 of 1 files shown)


TBBle:
Travis passed its checks for my commit and Pull Request, if that's what you mean. It's also how I use it locally, inside a Virtualenv environment on both Linux and Windows.

The instructions in the documentation on using this with Travis already rely on this, as they don't use `sudo`.

PR #1024, opened 2017-05-19
Don't leave non-standard "caniuse" attribute in output

### LINKED ISSUE #1023: remove caniuse attributes from the output
When **Include Can I Use Panels** is false (as it is for example on csswg WDs), caniuse attributes are kept in the output markup, which then cause the spec to fail validation when trying to publish with echidna.

https://labs.w3.org/echidna/api/status?id=8eefbfc8-9ac9-4950-bea9-6591b506b640

Even if the panels are not being generated, the caniuse attribute should never be present in the output.

frivoal:
Fixes #1023

### DIFF
--- a/bikeshed/caniuse.py
+++ b/bikeshed/caniuse.py
@@ -9,6 +9,9 @@ def addCanIUsePanels(doc):
     # Constructs "Can I Use panels" which show a compatibility data summary
     # for a term's feature.
     if not doc.md.includeCanIUsePanels:
+        elements = findAll("[caniuse]", doc)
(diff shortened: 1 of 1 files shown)


tabatkins:
I'm gonna reject this one; I try to avoid inflicting selector-matching on specs when I don't need to, as it's one of the largest timesinks in Bikeshed.  I'll instead do this cleanup in `cleanupHTML()`; thanks for the catch!

tabatkins:
Fix was committed in acdedda

PR #1025, opened 2017-05-19
Don't use unitless numbers in stroke-width property

frivoal:
Unitless numbers work in actual browsers, but they're not correct
according to the spec which requires a <length-percentage>,
and not doing so makes the W3C css validator complain and block
publication.


frivoal:
By the way, if unitless numbers work in all browsers, as seems to be the case, shouldn't we say so in the spec also?

PR #1027, opened 2017-05-22
Fix a Python warning

zcorpan:
When running bikeshed I got this warning:

```
__init__.py:1306: FutureWarning: The behavior of this method will change in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.
  if child and child.tag == "code":
```


tabatkins:
I really hate this warning. >_<  (Tho to be fair, the current way it works (truthy if len(el) is >0) means I *was* writing incorrect code.)

Issue #1038, opened 2017-06-06
backticks are still treated as special inside of opaque blocks


tabatkins:
That is, the following:

```html
<pre>
  let result = `foo${bar}`;
</pre>
```

...results in a pre containing `let result = <code>foo${bar}</code>;`, rather than the intended backticks for a template literal.

Other Markdown shorthands are not parsed inside of an opaque element, but backticks are done in a special early phase that doesn't look for opaque elements.

Showed up in the wild in <https://github.com/w3c/permissions/pull/149>

cbiesinger:
this also applies to \``` blocks, not just \<pre>

PR #1054, opened 2017-07-04
Correct processing of Markdown code spans & blocks

jugglinmike:
Ensure that the text content of "opaque" elements is not transformed.
Do not apply "code span" transformations to code blocks.

This is intended to resolve gh-1038.

### DIFF
--- a/bikeshed/__init__.py
+++ b/bikeshed/__init__.py
@@ -808,12 +808,12 @@ class MarkdownCodeSpans(func.Functor):
         for m in re.finditer(r"(\\`)|(`+)", text):
             if mode == "text":
                 if m.group(1):
-                    newText += text[indexSoFar:m.start()] + m.group(1)[1]
+                    newText += text[indexSoFar:m.end()]
                     indexSoFar = m.end()
                 elif m.group(2):
                     mode = "code"
                     newText += text[indexSoFar:m.start()]
-                    indexSoFar = m.end()
+                    indexSoFar = m.start()
                     escapeLen = len(m.group(2))
             elif mode == "code":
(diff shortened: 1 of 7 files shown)


tabatkins:
As far as I can tell, this ends up seriously regressing backtick support. Backticks need to be handled deeply in the markdown parser, because they negate *all* other markdown features inside of themselves.  My current backtick code is far from perfect, but it seems to do an okay job of handling most of this kind of thing, I think.

jugglinmike:
That certainly wasn't my intention! I thought that between the project's
existing tests and the tests added here, the implementation was sound. In light
of your feedback, I suspect I've missed something pretty fundamental. Could you
say more about how this patch regresses current support?

tabatkins:
Lol, three years later, sorry...

It regresses because this attempts to make inline-backticks be processed at the same time as the other markdown shorthands, which is, in particular, after HTML parsing. It would mean that `` `<foo>` `` no longer works, as HTML parsing will instead parse the `<foo>` as a start tag, and only much later will the shorthand code run and try to process the contents.

Also, due to the way my shorthand code is currently handled (which I'm *just about* to be done fixing, finally), shorthands are run in a particular order, where *all* instances of a particular type of shorthand are recognized and transformed, then all instances of the next, etc. The Markdown shorthands, for no particular reason, are handled near the end of that process. So by the time these run, most of the other shorthands would *also* have already had time to parse and transform, and thus not get caught by the backticks here.

Sigh, good call tho that my tests appear to be grossly inadequate here; they test that whitespace is preserved inside of backticks, but not that, say, HTML isn't parsed. Cool.

PR #1083, opened 2017-08-30
Allow to override metadata from command line in echidna mode

dontcallmedom:


tabatkins:
Heh, sorry about missing that one.

Issue #1086, opened 2017-09-01
Regression-test against all known Bikeshed specs


tabatkins:
My manually-written tests don't exercise all code paths, and I'm not optimistic they'll ever get there.  The variety of specs in the wild, however, pretty much do. I should be regression-testing against all the known specs.

foolip:
Do you already have a list of all specs using Bikeshed?

tabatkins:
Definitely not, I'm regularly surprised by where it gets used. But as a start we can grab all the csswg, Houdini, and whatwg specs, and can probably grep the W3C org for more .bs files.

xfq:
A quick search: https://github.com/search?utf8=%E2%9C%93&q=org%3Aw3c+extension%3Abs&type=Code

tabatkins:
Or here's the search minus all the CSSWG-related stuff, only 50 specs <https://github.com/search?utf8=%E2%9C%93&q=org%3Aw3c+extension%3Abs+-repo%3Aw3c%2Fcsswg-drafts+-repo%3Aw3c%2Ffxtf-drafts+-repo%3Aw3c%2Fcss-houdini-drafts&type=Code>

jfbastien:
[Hai](https://github.com/jfbastien/papers/tree/master/source)

foolip:
The branch https://github.com/foolip/bikeshed/tree/auto-test-update is now being updated daily by the scripts in https://github.com/foolip/bikeshed-tests. It has 197 .bs files.

One way of doing this might be to just merge all of that into Bikeshed weekly or so. Alternatively, Bikeshed's Travis job could pull in all of the .bs files and generate the HTML with and without the code changes in a PR, but then I don't know what should cause the job to fail, since some changes are correct.

@tabatkins?

tabatkins:
Manually merged in the stuff.  foolip, when you're grabbing the tests, are you *generating* the .html files, or *grabbing* them from the repos?

xfq:
The UI Events spec uses `<pre class=include>` to include the sections of the spec, but the `sections` directory isn't included in https://github.com/tabatkins/bikeshed/tree/master/tests/github/w3c/uievents

PR #1125, opened 2017-11-05
Support sudirectories in tests/

foolip:
This is to allow adding tests/whatwg/dom/dom.bs and similar.

Part of https://github.com/tabatkins/bikeshed/issues/1086.

### DIFF
--- a/bikeshed/test.py
+++ b/bikeshed/test.py
@@ -12,11 +12,22 @@ from .htmlhelpers import parseDocument, outerHTML, nodeIter, isElement, findAll
 from .messages import *
 
 
+TEST_DIR = os.path.abspath(os.path.join(config.scriptPath(), "..", "tests"))
+
+
 def findTestFiles():
-    for root, dirnames, filenames in os.walk(config.scriptPath() + "/../tests"):
+    for root, dirnames, filenames in os.walk(TEST_DIR):
         for filename in filenames:
             if filename.endswith(".bs"):
-                yield root + "/" + filename
+                yield os.path.join(root, filename)
+
+
+# The test name will be the path relative to the tests directory, or the path as
+# given if the test is outside of that directory.
+def testNameForPath(path):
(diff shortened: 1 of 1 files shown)


bikeshed/test.py
```diff
@@ -95,7 +101,7 @@ def equalOrEmpty(a, b):
 
 def rebase(files=None):
     if not files:
-        files = glob.glob(config.scriptPath() + "/../tests/*.bs")
+        files = findTestFiles()
     for path in files:
         _,_,name = path.rpartition("/")
         p("Rebasing {0}".format(name))
```
foolip: There's a problem here and with `p(testName)` above, in that just the filename is printed but the path relative to tests/ would be better. But how, given that the paths might also be outside? Use `os.path.abspath` on all paths and look for a prefix?

bikeshed/test.py
```diff
@@ -12,13 +12,19 @@
 from .messages import *
 
 
+def findTestFiles():
+    for root, dirnames, filenames in os.walk(config.scriptPath() + "/../tests"):
+        for filename in filenames:
+            if filename.endswith(".bs"):
+                yield root + "/" + filename
```
tabatkins: Mind using os.path.join() here?

And while we're here, swap out the os.walk argument to use the path functions too.

bikeshed/test.py
tabatkins: Yeah, removing a common prefix is pretty easy.

bikeshed/test.py
```diff
@@ -12,18 +12,35 @@
 from .messages import *
 
 
+TEST_DIR = os.path.abspath(os.path.join(config.scriptPath(), "..", "tests"))
```
foolip: Just `os.path.abspath(config.scriptPath() + "/../tests")` would probably have worked as well because of abspath, but I went for eliminating path separators.

foolip:
Ping @tabatkins

PR #1133, opened 2017-11-09
Patch from hell: Preliminary (non-working) Python 3 support.

cynthia:
**Nota bene: Do not merge, breaking changes.**

Figured this would be worth sending in first, then iterating through nicer ways through code review before I forget that this patch existed in the first place.

The plug-in loading code does not work in Py3, I'm suspecting it is due to a change in the class reference behavior between Py2 and Py3.

(The patch was mostly written during plenary day's bikeshed session, so mind the sloppiness.)


jyasskin:
Looks like this one's been fixed! 🎉

PR #1140, opened 2017-11-20
Clarify the effects of `.example`

harbichidian:
While notes, issues, and examples are mentioned in the same heading, it is not clear what results you get by adding the `.example` class to an element.


harbichidian:
Will update with recompiled `.html`, shortly

xfq:
I think preserving the [semantic linefeeds](http://rhodesmill.org/brandon/2012/one-sentence-per-line/) might be preferred. It makes the reviewer's life easier :)

PR #1152, opened 2017-12-13
Automatic test update

foolip:
Using bikeshed-tests commit bfd2edfbdbae4b9adfeeba65ce0ae097d0912f59

https://travis-ci.org/foolip/bikeshed-tests/builds/315676782


foolip:
This commit was created automatically, but this PR opened manually. Please read through the Travis log for a lot of weird things, not sure which should be fixed.

Part of https://github.com/tabatkins/bikeshed/issues/1086.

foolip:
In the log, expand the "find tests/github -type f | while read file; do" bit to find the interesting stuff. I should probably move that to before deploy so it's not collapsed by default.

foolip:
As expected, fixing stuff and letting Travis run updates this PR. A possible automated workflow, then, is to always create a PR if one doesn't exist, and otherwise updating. @tabatkins, then you can merge those PRs on some cadence that makes sense, weekly or before some changes are about to happen.

tabatkins:
Varying stuff is:
* [ ] Bikeshed's commit sha
* [ ] Generation date

Current specs are all part of `Group: test`, which fixes the date and turns off the commit sha via default metadata. Since these new specs aren't part of that Group, they don't get that treatment. I need to copy over the "metadata from the command line" feature from the plain `spec` command to also work when rebaselining tests.

tabatkins:
All right, closing because I merged this manually. Bikeshed's testing framework now automatically kills the varying things, and the comparison now just uses the output directly, no complicated munging that can hide bugs.

tabatkins:
(In 70a0a014)

foolip:
Cool, will see about auto-updating the files too.

PR #1155, opened 2017-12-18
Update WHATWG boilerplate

domenic:
This uses the computed metadata features to set better defaults across all WHATWG standards, and moves certain scripts (including the service worker registration script) into the boilerplate.

As a drive-by, it removes the "likely out-of-date" for translations as it turns out we have very responsive translators for WHATWG standards and we'd like to be a bit nicer about their work.

### DIFF
--- a/bikeshed/boilerplate.py
+++ b/bikeshed/boilerplate.py
@@ -752,7 +752,7 @@ def addSpecMetadataSection(doc):
             # so this prevents code from genning an extra wrapper <dd>.
             return [E.dt({"class": "editor"}, displayKey, ":")] + vals
         elif key == "Translations":
-            ret = [E.dt(displayKey, " ", E.small("(non-normative and likely out-of-date)"), ":")]
+            ret = [E.dt(displayKey, " ", E.small("(non-normative)"), ":")]
             ret += [E.dd({}, val) for val in vals]
             return ret
         else:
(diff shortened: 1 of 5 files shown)


domenic:
Please DO NOT MERGE until we get https://github.com/whatwg/sg/issues/51 resolved, as otherwise all WHATWG spec builds will start failing. The plan was to land this then land a bunch of PRs similar to whatwg/url#360 in quick succession. But until whatwg/sg#51 is resolved those PRs are blocked, and so this one is too.

I'll ping the thread once it's ready!

domenic:
@annevk I think this is ready. Let me know if it looks good to you; if so I'll send PRs like this to every other Bikeshed-using WHATWG spec, and then we can get the Bikeshed PR merged, and everything will be nice.

domenic:
Er, I meant to actually ping you to take a look at https://github.com/whatwg/url/pull/360, @annevk. Sorry about that.

domenic:
I got alphabetically down to Quirks before being called away to dinner. Quirks has some small issues as well: https://github.com/whatwg/quirks/pull/20. Storage, Streams, URL, and XHR remain.

domenic:
All PRs are ready. Things that would be nice to fix, but will not completely break things if they are done later:

- #1161
- https://github.com/whatwg/whatwg.org/pull/155

domenic:
OK, merge at will!

PR #1188, opened 2018-02-14
Update WebAssembly abstract + status to have correct verbiage.

flagxor:


bikeshed/boilerplate/wasm/abstract.include
```diff
@@ -2,6 +2,6 @@
 
 [ABSTRACT]
 Part of a collection of related documents:
-the <a href="https://www.w3.org/TR/wasm-core/">Core WebAssembly Specification</a>,
-the <a href="https://www.w3.org/TR/wasm-js-api/">WebAssembly JS Interface</a>,
-and the <a href="https://www.w3.org/TR/wasm-web-api/">WebAssembly Web API</a>.
+the <a href="https://www.w3.org/TR/wasm-core-1/">Core WebAssembly Specification</a>,
```
domenic: Nooo don't link to the numbered specs...

flagxor: These should be the un-numbered one?

domenic: I mean, yeah, ideally people should not ever look at the numbered specs, I think. (I am assuming this text is under your control, not part of some process-mandated boilerplate, since I haven't seen similar boilerplate before.)

flagxor: For those particular links going versionless makes sense, so I've changed them.
In general, this is totally me trying to fit the process :-)

jfbastien:
Seems kinda weird to modify bikeshed to talk about first public working draft? Can't that be in the document's intro, or some metadata that gets auto-included based on something in the `<pre>` headers?

tabatkins:
Huh? This commit updates the wasm group's own boilerplate. Bikeshed just holds a lot of boilerplates for convenience, so you don't have to carry it around yourself.

jfbastien:
Yeah I just mean that it's odd to have that in the boilerplate, instead of the source doc itself. 🤷‍♂️

PR #1210, opened 2018-03-26
Add boilerplate for WG14 (the C language)

jfbastien:
As-is it doesn't support paper numbers. Their URL scheme seems simple enough: it's all at http://www.open-std.org/jtc1/sc22/wg14/www/docs/


xfq:
Why is the default highlighting language C++?

C seems to be supported, according to http://pygments.org/docs/lexers/#lexers-for-c-c-languages

jfbastien:
> Why is the default highlighting language C++?

Because I copy pasta'd from the C++ boilerplate 😁 
Fixed here: https://github.com/tabatkins/bikeshed/pull/1215

PR #1218, opened 2018-04-05
Automatic test update

foolip:
By https://github.com/foolip/bikeshed-tests


foolip:
Yay! This is the first PR created by Travis, and as when it's closed another should be opened. @tabatkins, can you merge this to see if that in fact works? If so, then our work here is done, PRs can be merged at your convenience. Note that the PR description isn't exactly the same as the commit message and the commit message is updated every day to point to the build that produced it, so doing rebase and merge instead of squash and merge is probably best. I can change any of the strings as you like too.

foolip:
This PR now updated after https://github.com/foolip/bikeshed-tests/pull/4, not broken still :)

PR #1238, opened 2018-04-28
Assuming utf8 in and out for inlineTags.

flagxor:
Maybe like this?


tabatkins:
`.encode()` and `.decode()` are actually built into the `unicode` and `str` types already. Fixed!

PR #1239, opened 2018-04-28
WIP Boilerplate for WHATWG Review Drafts

domenic:
/cc @annevk 

Preview: https://rawgit.com/domenic/a1cc79bd89d2610f096bc29dfbab52d2/raw/3656877ce9811394aaf681ab48b7b1ca011dcb62/dom-review-draft-generated.html

Things to do:

- [ ] Refactor CSS to reduce duplication with the existing files. Right now it's all inline in the header. (Maybe merge bikeshed.css into standard.css while we're at it...)

Optional things to do that may require extra Bikeshed features; thoughts welcome @tabatkins:

- Remove generated .self-links
- Remove syntax highlighting (not just the styles; the various spans it inserts too)
- Remove data-dfn-for/data-dfn-type/data-export/data-link-type etc. attributes. This document should never be ingested by Shepherd or other tools.
- Remove the various IDs like ` id="ref-for-concept-tree-child⑤"` that are not used now that we turn off dfn panels


tabatkins:
I'm happy to add some metadata that either prevents or removes all those things.

domenic:
Nice, the tests found that I broke non-LS things like loader by rearranging the metadata... let me try some fixes.

domenic:
From my perspective this is good to go. @annevk, want to check over this and https://github.com/whatwg/whatwg.org/pull/200? @tabatkins, mind doing a quick check on our usage of Bikeshed boilerplate/defaults/etc.?

tabatkins:
Looks good to me! Note my response in IRC about requiring a Date - if you add RD to the list of snapshotStatuses, it'll make `Date` recommended (which then generates a warning).

(Anne asked a while back about something that turns *all* warnings/errors into fatal errors. Want me to implement that so y'all can turn it on for these drafts?)

domenic:
I think we want that eventually but we need to audit all our stuff and get rid of the warnings first; it's close but not quite there. So, yes, but no rush.

bikeshed/boilerplate/whatwg/header-RD.include
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE html>
+<html lang="en">
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+<meta name="theme-color" content="#3c790a">
```
annevk: Maybe we shouldn't do this, given they are generally unstyled? Same for the favicon.

bikeshed/boilerplate/whatwg/header-RD.include
```diff
@@ -0,0 +1,43 @@
...
+      text that is identified as "Objection Pending" or "Confirmation Pending". Developers should
+      refer to the <a href="https://[SHORTNAME].spec.whatwg.org/">Living Standard</a> for the most
+      current error corrections and other developments.
+    <p>
+      For information regarding patent commitments, please see the
+      <a href="https://whatwg.org/ipr-policy">IPR Policy</a> and
+      <a href="https://github.com/whatwg/[SHORTNAME]/labels/exclusion%20notices">exclusion notices</a>.
+
+    <p>
+      Do not attempt to implement this version of the specification. Do not reference this version
```
annevk: Can we change this to "standard"?

bikeshed/config/status.py
```diff
@@ -77,13 +78,14 @@
     "fido/PS": "Proposed Standard",
     "fido/FD": "Final Document"
 }
-snapshotStatuses = ["w3c/WD", "w3c/FPWD", "w3c/LCWD", "w3c/CR", "w3c/PR", "w3c/REC", "w3c/PER", "w3c/WG-NOTE", "w3c/IG-NOTE", "w3c/NOTE", "w3c/MO"]
-unlevelledStatuses = ["LS", "LD", "DREAM", "w3c/UD", "LS-COMMIT", "LS-BRANCH", "FINDING", "DRAFT-FINDING"]
+snapshotStatuses = ["w3c/WD", "w3c/FPWD", "w3c/LCWD", "w3c/CR", "w3c/PR", "w3c/REC", "w3c/PER", "w3c/WG-NOTE", "w3c/IG-NOTE", "w3c/NOTE", "w3c/MO", "whatwg/RD"]
+unlevelledStatuses = ["LS", "LD", "DREAM", "w3c/UD", "LS-COMMIT", "LS-BRANCH", "FINDING", "DRAFT-FINDING", "whatwg/RD"]
 deadlineStatuses = ["w3c/LCWD", "w3c/PR"]
-noEDStatuses = ["LS", "LS-COMMIT", "LS-BRANCH", "LD", "FINDING", "DRAFT-FINDING", "DREAM", "iso/NP"]
+noEDStatuses = ["LS", "LS-COMMIT", "LS-BRANCH", "LD", "FINDING", "DRAFT-FINDING", "DREAM", "iso/NP", "whatwg/RD"]
```
annevk: I don't understand why LS is not prefixed, but this one is. And why is LS-COMMIT and LS-BRANCH not a snapshot status?

tabatkins: LS (and related) aren't associated with any particular mega-group, but your Review Draft status is definitely whatwg-specific, so it's tagged accordingly.

Why they're not snapshot statuses is a good question, they probably should be.

domenic: Snapshot statuses would require them to have a `Date` manually added, right? Whereas it's reasonable enough to have auto-generated dates for such snapshots, since they're only generated once and never updated.

bikeshed/boilerplate/whatwg/header-RD.include
domenic: I think there's a difference between styling and branding; I think it makes sense to keep the branding?

annevk: I think I'd prefer it if we used a different favicon (and maybe color if we want to keep that) for these so you can tell from your tabs which are Living and which are stale.

annevk: We should also consider that for snapshots I suppose.

domenic: Good idea! Maybe grayed out? Or more extreme, upside down and inverted?

annevk: I like grayed out.

domenic: Updated (depends on https://github.com/whatwg/whatwg.org/pull/204). I'd like to do this for commit snapshots too but that's a tiny bit trickier and we already have a lot of things in flight. Let me file a meta issue to remind us and we can do it later.

bikeshed/boilerplate/whatwg/header-RD.include
```diff
@@ -0,0 +1,43 @@
...
+<link href="https://resources.whatwg.org/spec.css" rel="stylesheet">
+<link href="https://resources.whatwg.org/review-draft.css" rel="stylesheet">
+<link href="[LOGO]" rel="icon">
+
+<body class="h-entry status-[STATUS]">
+<div class="head">
+  <p data-fill-with="logo"></p>
+  <hgroup>
+    <h1 id="title" class="p-name no-ref">[SPECTITLE]</h1>
+    <h2 id="subtitle" class="no-num no-toc no-ref">[LONGSTATUS] — Last Updated <time class="dt-updated" datetime="[ISODATE]">[DATE]</time></h2>
```
annevk: Should we say "Published" here since we don't intend to revise the date when we do an update?

bikeshed/boilerplate/whatwg/computed-metadata-RD.include
```diff
@@ -0,0 +1,4 @@
+{
+  "Title": "[SPECTITLE] Standard Review Draft [ISODATE]",
```
annevk: It seems somewhat friendlier to use "May 2018" in the title.

domenic: @tabatkins can we get a macro that turns the defined Date: metadata into a human-readable "May 2018" format?

tabatkins: Done. Didn't know whether you wanted short or full month names, so I added both - `[DATE-MY]` gives short month, `[DATE-MMY]` gives full month.

(And for consistency, `[DATE]` now has `[DATE-DMMY]` as an alias.)

bikeshed/boilerplate/warning-commit.include
```diff
@@ -5,7 +5,7 @@
     and should only be used as a historical reference.
     This commit may not even have been merged into master.
```
annevk: As per comments against HTML we should remove this line.

bikeshed/boilerplate/whatwg/header-RD.include
```diff
@@ -0,0 +1,43 @@
...
+
+<body class="h-entry status-[STATUS]">
+<div class="head">
+  <p data-fill-with="logo"></p>
+  <hgroup>
+    <h1 id="title" class="p-name no-ref">[SPECTITLE]</h1>
+    <h2 id="subtitle" class="no-num no-toc no-ref">[LONGSTATUS] — Published <time class="dt-updated" datetime="[ISODATE]">[DATE]</time></h2>
+  </hgroup>
+  <details class="annoying-warning" open>
+    <summary>This Is a Review Draft of the Standard</summary>
```
annevk: This is also wrong for the other warnings. We shouldn't use titlecase here. This should be "This is a Review Draft of the standard".

PR #1270, opened 2018-06-11
Default favicon for wg21 added

mpusz:
Added default C++ favicon for WG21 documents


jfbastien:
Some people in the C++ committee have complained before about papers containing any kind of external link, partly because they're worried about tracking. This seems pretty harmless because it's a favicon (which browsers auto-fetch when they can) and it's hosted by isocpp.

Maybe we could base64 encode it? Something like https://stackoverflow.com/questions/5199902/isnt-it-silly-that-a-tiny-favicon-requires-yet-another-http-request-how-to-mak

tabatkins:
It's a non-issue. As you said, if there's no favicon at all, the browser *still makes a request for one*, guessing at the likely URL. So this doesn't increase the number of requests made, it just changes where they point to.

And having it be a proper favicon means it's cached better than inlining it. A small concern, since favicons are tiny, but still.

PR #1273, opened 2018-06-12
Fix the example header.include file

foolip:


foolip:
A few problems all fixed at once:
* Double abstract heading would result from original code
* [TITLE] and other things expanded, making https://tabatkins.github.io/bikeshed/#bp-new not useful
* Missing `<main>`

foolip:
@tabatkins, is https://tabatkins.github.io/bikeshed/#bp-new not auto-updated?

tabatkins:
Nope 😀

PR #1287, opened 2018-06-22
Don't include dfn panel styles in WHATWG RDs

domenic:
DFN panels are not included anyway.

(I wonder if "Use Dfn Panels: No" should do this automatically?)


annevk:
This is closed by #1767. Even if "use dfn panels" turns it off I think it's still worth including in this case so this line is no different from the non-RD version.

Issue #1306, opened 2018-07-11
Inconsistencies relating to the H1 metadata field


brycelelbach:
- [ ] 0) [The documentation refers to an `[H1]` text macro](https://tabatkins.github.io/bikeshed/#text-macros), but the actual name of the text macro seems to be `[SPECTITLE]`. There are three options here: (a) make the names consistent by renaming the `[SPECTITLE]` metadata field to `[H1]` and updating the docs, (b) make the names consistent by renaming the H1 metadata field to Spectitle and updating the docs, or (c) leave the names inconsistent, and update the docs to refer to `[SPECTITLE]` instead of `[H1]` as the text macro name. Which do you think makes sense? I favor (a).
- [ ] 1) The boilerplates for all specs use the `[TITLE]` text macro for both the `<title>` element and the `<h1>` element. Pull request #1305 changes the WG14 and WG21 boilerplates to use `[SPECTITLE]` for the `<h1>` element. I'm happy to send a separate pull request for all the other boilerplates, if that makes sense.
- [ ] 2) (Minor) [In bullet point describing the Title metadata field and the note describing the H1 metadata field just below it](https://tabatkins.github.io/bikeshed/#metadata), there are references to `<title>` and `<h1>`, both of which use autolinking. The `<title>` reference correctly has a link to the HTML spec, but the `<h1>` reference does not.

tabatkins:
I do need to fix the documentation inconsistency; I'm not sure why I did it that way. But `[SPECTITLE]` is already in use elsewhere (in the WHATWG boilerplate, at least) and I try not to make breaking changes without good reason; so I should just change what's in the docs.

PR #1307, opened 2018-07-11
Rename the `[SPECTITLE]` text macro to `[H1]` and use it in the WG14 and WG21 boilerplate.

brycelelbach:
* Rename the `[SPECTITLE]` macro to `[H1]`, to match the metadata field and the docs.
* In the `<h1>` in the WG14 and WG21 boilerplate, use the `[H1]` macro.

See #1306 for background. This PR assumes that the inconsistency in naming between the H1 metadata field and the `[SPECDATA]` text macro is a mistake, and renames the `[SPECDATA]` to `[H1]`, which is what the Bikeshed documentation says it should be called.

**Either this Pull Request or #1305 should be taken, but not both; they are mutually exclusive.**

### DIFF
--- a/bikeshed/boilerplate/wg14/header.include
+++ b/bikeshed/boilerplate/wg14/header.include
@@ -25,7 +25,7 @@
 <body class="h-entry">
 <div class="head">
   <p data-fill-with="logo"></p>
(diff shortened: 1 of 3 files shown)


brycelelbach:
Hey Tab ping?

tabatkins:
Merged #1305, so closing this.

PR #1309, opened 2018-07-11
Add a `computed-metadata.include` for WG21

brycelelbach:
* Allows the URL metadata field to be omitted - if it is, the link is computed from the paper and revision number using wg21.link.

### DIFF
--- a/bikeshed/boilerplate/wg21/computed-metadata.include
+++ b/bikeshed/boilerplate/wg21/computed-metadata.include
@@ -1,0 +1,3 @@
+{
+    "URL": "https://wg21.link/[SHORTNAME]R[LEVEL]"
+}


jfbastien:
Oooh nice!

tabatkins:
Note that this isn't "if omitted", it's "always" - computed-metadata has a higher precedence than in-document metadata.

(Another recent bug requested a way to do higher-priority in-document metadata for when you need computed-metadata for most things but *occasionally* want to override it, which I'll be adding soonish.)

tabatkins:
ping @brycelelbach !

brycelelbach:
Hrm. @jfbastien what do you think?

I'm inclined to say it's fine and we should go ahead and merge it.

jfbastien:
Yeah that works, it’s always the right value, even for a draft. Can you make sure WG14 doesn’t get the same thing though?

tabatkins:
Are y'all still interested in this, or should I close it?

jfbastien:
I think Bryce should do it.

PR #1312, opened 2018-07-11
Use crossorigin="" attribute in WHATWG templates

domenic:
See https://github.com/whatwg/whatwg.org/issues/211 for discussion.

We probably shouldn't merge this until people review https://github.com/whatwg/whatwg.org/pull/202; this whole area is a bit weird and I want to make sure more people than just me think this is the right direction.

### DIFF
--- a/bikeshed/boilerplate/whatwg/header-RD.include
+++ b/bikeshed/boilerplate/whatwg/header-RD.include
@@ -4,8 +4,8 @@
 <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
 <meta name="theme-color" content="#3c790a">
 <title>[TITLE]</title>
-<link href="https://resources.whatwg.org/spec.css" rel="stylesheet">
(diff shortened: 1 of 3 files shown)


tabatkins:
Since <https://github.com/whatwg/whatwg.org/pull/202> is merged, should I merge this?

domenic:
So, we should probably merge this. But, would you be interested instead/in addition in adding a feature to Bikeshed that adds `crossorigin=""` to all link/img/script/audio/video elements automatically, i.e. opts you out of legacy SOP exceptions?

tabatkins:
I'm totally fine with that. I'll commit a new metadata for it in a few minutes.

tabatkins:
[All right, added.](https://tabatkins.github.io/bikeshed/#metadata-force-crossorigin) Should I go ahead and close this PR so you can instead just switch to using `Force Crossorigin` in WHATWG defaults?

domenic:
Yeah, sounds like a plan!

Issue #788, opened 2016-08-18
Literal backticks.


mikewest:
There doesn't seem to be a way to embed a literal `` ` `` in a Bikeshed document. WHATWG specs use backticks to demarcate byte-sequences from strings, so it would be nice to support this use case. We should probably support escapes, a la http://spec.commonmark.org/0.26/#example-287.

tabatkins:
Okay, I rewrote backtick handling to allow for escapes.  Should work now.

mikewest:
You == great. Thanks!

PR #1315, opened 2018-07-16
Show the inline markdown sequences.

jyasskin:
Including backtick escapes from #788.


jfbastien:
Maybe mention the ones from <marquee> and <blink>?

jyasskin:
@jfbastien Follow-up patches welcome.

PR #1340, opened 2018-08-28
Use the new "Force Crossorigin" option for WHATWG specs

domenic:
Also consistently use "yes" and "no" instead of "true" and "false".


annevk:
This broke the Encoding Standard and presumably other standards?

https://github.com/whatwg/encoding/pull/152. Adding it to `a` elements is incorrect.

tabatkins:
Sorry about that, fixed.

PR #1350, opened 2018-09-23
allow webaudio WG to generate CR from bikeshed

svgeesus:
Status boilerplate for CR, copied from csswg CR boilerplate with WG name, feedback, and patent links updated from webaudio WD boilerplate.


tabatkins:
Filename is `.include.txt`, it needs to be just `.include`

PR #1362, opened 2018-10-05
Logo the right way round (for LTR languages)

Smylers:
While looking at a CSS spec I saw an unfamiliar logo in the corner. Stopping to look at it, it appeared to be a shed with a bike coming out of it. “Shed. Bike. Shed-bike? Oh, I think I've heard of a spec-generation tool called Bikeshed.”

Given you are called Bikeshed and not Shedbike, flipping your logo horizontally would make it more intuitive for people encountering it for the first time.

Well, at least for people used to reading left to right. But given the word ‘bikeshed’ is English and needs to be read left to right, you're already requiring that.


tabatkins:
The icon isn't meant to be a rebus? It's just... a bikeshed. A shed containing a bike. Bikesheds point in both directions.

PR #1374, opened 2018-10-22
WG21: add convenience mapping

jfbastien:
Maybe we should map all of the convenience names to their actual SG name? These two seem pretty neat to have.

Or maybe we should allow either convenience name or SG name, and print out both? i.e. EWGI gets you "SG17: EWGI"?


jfbastien:
@brycelelbach what do you think? It's confusing with all the SG numbers...
https://isocpp.org/std/the-committee

brycelelbach:
+1

jfbastien:
Ping ping.

PR #1379, opened 2018-11-01
WG21: More pseudonyms

jfbastien:


tabatkins:
Do you want to require that specific capitalization?

jfbastien:
Not necessarily, but I don't really see people capitalizing them differently either. @brycelelbach WDYT?

PR #1387, opened 2018-11-08
Boilerplate for FXTF specs in csswg space

dirkschulze:
Don't know if this is the best or even right thing to do. I created a boilerplate for FXTF specs that do not live in the fxtf-drafts repo but in the csswg-drafts repo.


dirkschulze:
CC @birtles and @tabatkins

PR #1391, opened 2018-11-13
Create status-NOTE.include

svgeesus:
Status was missing for notes, so was not being generated. Notice changes in 2018, like `data-deliverer`.


svgeesus:
Uh, this is for CSSWG, I should have said in the original commit message.

PR #1392, opened 2018-11-13
Add "proposed" attribute for IDL

szager-chromium:
When an IDL declaration has the 'proposed' extended attribute (i.e.,
it begins with the literal text '[proposed]'), then the generated
html for both the declaration and the definition will be enclosed in
an element with class='proposed'.

### DIFF
--- a/bikeshed/idl.py
+++ b/bikeshed/idl.py
@@ -54,6 +54,8 @@ class IDLMarker(object):
     def markupConstruct(self, text, construct):
         # Fires for every 'construct' in the WebIDL.
         # Some things are "productions", not "constructs".
+        if 'proposed' in construct.extendedAttributes:
+          return ("<span class='proposed'>", "</span>")
         return (None, None)
 
     def markupType(self, text, construct):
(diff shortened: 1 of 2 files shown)


domenic:
This seems like a bad idea given there's no Web IDL spec  for such an attribute.

Instead, you could just wrap the IDL with such spans yourself.

szager-chromium:
I have wrestled quite a bit with adding those spans and getting bikeshed to do the right thing; some constructs work, others don't.

Here's an example spec produced with this patch:

http://szager-chromium.github.io/IntersectionObserver/

I tried to produce identical styling without hand-editing the output or patching bikeshed, but I couldn't.

tabatkins:
Yeah, Bikeshed doesn't generally allow annotating a WebIDL block with markup; it generates its own markup (for `a` and `dfn`) and merging two markup trees isn't possible in general. (I can do it for highlighting because highlighting is, in effect, providing color to individual characters; it uses spans containing multiple characters purely for optimization purposes. Thus I can always push highlighting markup "inside of" other markup without breaking anything.)

That said, I could always *remove* this particular extended attribute from the WebIDL before rendering it, and just reflect it in a wrapper with a class instead.

Alternately, we *could* just add it to WebIDL as a no-op, or even a recognized but guaranteed invalid extended attribute.

domenic:
Stripping before outputting makes sense to me. That does mean Bikeshed has its own Web IDL dialect it takes as input, but maybe that's not so bad, if it's well-documented.

Alternately, it may work to just change the syntax to `/* proposed */` instead of `[proposed]`

PR #1409, opened 2018-12-13
Allow manually specifying the snapshot URL

frivoal:
This comes handy if you want to use bikeshed with one of the snapshotStatues to produce something that looks exactly like one of the things that goes on TR, without actually going on TR. For instance, the Process (or the patent policy, or any number of official dated documents that do or should have an ED as well).

In addition, creating a new status (accessible to anyone?) within datedStatuses group would also be useful, so that I didn't have to use header.include to suppress the insertion of `<h2>[LONGSTATUS]</h2>` would also be nice, but:
* That's an addition, not an alternative to this patch, so you should still take this in
* for now, I can indeed use header.include and to abuse another Status (like REC), so that's not a pressing need
* Creating a new status means I need to pick a name for it, and naming is hard :)

### DIFF
--- a/bikeshed/boilerplate.py
+++ b/bikeshed/boilerplate.py
@@ -731,7 +731,7 @@ def addSpecMetadataSection(doc):
         md["This version"].append(E.a({"href":mac['version'], "class":"u-url"}, mac['version']))
     if doc.md.TR:
         md["Latest published version"].append(E.a({"href": doc.md.TR}, doc.md.TR))
-    if doc.md.ED and doc.md.status in config.snapshotStatuses:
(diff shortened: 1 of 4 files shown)


frivoal:
I still need this. Can you let me know if it's ok, or if I need to change something?

tabatkins:
I was concerned about the possible effect this might have on WHATWG documents, since I think it would start hitting them. But I haven't tested yet, so I'm not sure. Will review more thoroly soon.

KimiSon:
I'll just report this to the police then

frivoal:
Ping. Process 2020 and Patent Policy 2020 are wrapping up, and I'll be needing this very soon.

Issue #1416, opened 2018-12-29
WPT errors shouldn't be fatal


fantasai:
Making broken or missing WPT tests a fatal error means the spec stops building due to changes in the WPT repository, which isn't acceptable.

(It also means I have to spend a bunch of time stripping out the WPT annotations of specs which have incomplete annotations in order to publish, which is a waste of time.)

koto:
+1, especially given that the test list is hardcoded in the file in bikeshed (which is outdated compared to the current WPT list by a couple of months). This is especially cumbersome for new specs & implementations that actively work on the test files.

tabatkins:
It's absolutely not hardcoded, it's picked up as part of the update process like the definition files.

koto:
Oh, I totally missed that `bikeshed update` exists, sorry!

tabatkins:
Heh, no problem. I need to add a staleness check that complains at you if the data files are older than a week or two.

fantasai:
OK, but can we still get this fixed? Even with up-to-date indexes, having changes in the WPT repo break spec builds is still a problem, and I don't think it makes sense for these errors to be fatal.

tabatkins:
That's very much intentional, tho. If you're using WPT, keeping the spec and the tests healthy and aligned is very important. It's not something you should be ignoring.

tabatkins:
I'm not sure why you think you need to immediately do anything. That's what the `-f` flag is for, after all. And fatal errors don't stop the csswg server from building specs, either.

mkruisselbrink:
FWIW I completely agree here that it seems insane that the only way to work around this is by ignoring all fatal errors. This really doesn't seem like it warrants being treated more strictly than any other link errors.

tabatkins:
I still think link errors *should* be fatal, honestly; they harm the readability of the spec. The reason I relented on that issue, and am resisting on this one, is because of (a) where failures can come from, and (b) how the spec reads when failures happen.

For linking errors, failures can come from *anywhere*. A failure caused by the target definition changing or being removed is important to fix, of course, but failures can also come from *new definitions being added somewhere else* that just happen to make your autolink ambiguous now when it wasn't before. On the other hand, WPT errors can happen for precisely one reason: one of the tests for your spec was moved/renamed/deleted in the WPT repo. Easy to find, easy to fix, and generally rare; plus, as an editor you should be aware of changes to your spec's test folder.

[...]

PR #1417, opened 2018-12-29
Make WPT errors triggered by changes in WPT repo non-fatal

fantasai:
Related to issue #1416.


tabatkins:
Closing, since per the thread in #1416 I did switch the one significant error that doesn't harm readability to be a warning rather than a fatal.

PR #1427, opened 2019-02-02
Added support including SVG diagrams.

### LINKED ISSUE #1403: Bikeshed thinks empty lines in the middle of an svg start a new paragraph
There seems to be a weird parsing bug in Bikeshed that corrupts svg elements when there's empty lines in the, as it seems to thing that these are new paragraphs.

Try to put in an empty document the SVG illustration found in section https://w3c.github.io/w3process/#recs-and-notes of the W3C process, and see what bikshed does with it.

Removing empty lines helps. That shouldn't be needed. If the SVG is well formed, bikeshed should just leave it alone and not try to interpret the content. (and if it is not well formed, it should just fatal error).

NicolasRouquette:
Fixes #1403; see the new clause 5.11.2 in the bikeshed specification.


NicolasRouquette:
@tabatkins I rebased this branch onto master and re-generated doc/index.html to resolve the earlier conflict. Can you merge this PR?

PR #1434, opened 2019-02-18
Add Khronos Group and boilerplate for Khronos WebGL

Richard-Yunchao:
PTAL. Thank you very much. 

Khronos WebGL WG tend to use bikeshed to write its specification for WebGL 2 compute. But Khronos Group (https://www.khronos.org/) is not included in bikeshed. So I simply added Khronos Group and a boilerplate for Khronos WebGL Working Group, in order to support its own copyright and other things. Then my change to add WebGL 2 compute specification (https://github.com/KhronosGroup/WebGL/pull/2769) for Khronos WebGL Working Group can work. 

I am a new guy for bikeshed project. If this is not the correct way or if there are any other concerns, please let me know. Thank you. 

FYI, @kenrussell, @gyagp.

### DIFF
--- a/bikeshed/boilerplate/webgl/status.include
+++ b/bikeshed/boilerplate/webgl/status.include
@@ -1,0 +1,3 @@
+<p>This document is an editor's draft. Do not cite this document as other than work in progress.</p>
+
+<p>[STATUSTEXT]</p>
--- a/bikeshed/boilerplate/webgl/copyright.include
+++ b/bikeshed/boilerplate/webgl/copyright.include
@@ -1,0 +1,1 @@
+<a href="https://www.khronos.org/registry/speccopyright.html">Copyright</a> © [YEAR] the Contributors to the [TITLE] Specification, published by the <a href="https://www.khronos.org/webgl/wiki/Main_Page">Khronos WebGL Working Group</a> under the <a href="https://www.khronos.org/registry/speccopyright.html">Khronos Group License Agreement</a>.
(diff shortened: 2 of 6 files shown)


bikeshed/boilerplate/webgl/status.include
```diff
@@ -0,0 +1,3 @@
+<p>This document is an editor's draft. Do not cite this document as other than work in progress.</p>
+
+<p>[STATUSTEXT]</p>
```
kenrussell: Does this status.include properly handle the case where the specification is promoted from draft to official?

Richard-Yunchao: @kenrussell , I suppose that you want to format multiple Khronos WebGL related specifications via bikeshed, and these specifications are at different status (working draft, editor draft, ratified/offical, etc). 

Bikeshed can meet this requirement. We only need to submit another pr to add multiple status.include files. For example, rename this one to status-ED.include for editor's draft, and add status-WD.status for working status, etc.

In addtion, we need to add all other status like working draft and ratified into bikeshed/config/status.py. You can see that I have already added khronos/ED for editor's draft in that file.  

You see, it's quite easy and straightforward. And I verified the method above on my local machine.

kenrussell:
Thanks Yunchao. We've been planning to move the WebGL 1.0 and 2.0 specs to Bikeshed for quite some time and appreciate your pioneering this with the WebGL 2.0 Compute spec.

CC @grorg @jdashg

PR #1446, opened 2019-03-07
Get rid of <dfn> for "processing equivalence" in webplatform footer.

mkruisselbrink:
Having this as a `<dfn>` just seems to result in warnings when you're not actually
referencing it from the spec. Footers for other groups don't have this as a dfn either,
so just don't make it a `<dfn>`.

### DIFF
--- a/bikeshed/boilerplate/webplatform/footer.include
+++ b/bikeshed/boilerplate/webplatform/footer.include
@@ -37,7 +37,7 @@
     "should", "may", etc) used in introducing the algorithm.</p>
 
     <p>Conformance requirements phrased as algorithms or specific steps can be
-    implemented in any manner, so long as the end result is <dfn lt="processing equivalence" id="dfn-processing-equivalence">equivalent</dfn>. In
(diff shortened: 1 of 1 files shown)


mkruisselbrink:
Not sure if there is an easy way to verify if any of the webplatform specs actually do reference this definition... I can't imagine what they would do that for though.

mkruisselbrink:
@tabatkins ping?

tabatkins:
Happy to merge if you are. There's not currently any easy way to tell where a definition is used; it's been on my todo list for a while.

I agree tho that it's *exceedingly* unlikely anyone would be relying on it.

mkruisselbrink:
Yeah, I don't know how we'd check if this is used either. This isn't even "check where this one global definition is used" (which would be great to have), but this is "check which specs use the webplatform boilerplate, and use their own copy of this defintion". I.e. every spec would have their own unique definition here... 

Not sure why I even added it when originally contributing this boilerplate, probably just copied from some other footer.

PR #1453, opened 2019-04-13
Phrasing tweak in the documentation

frivoal:
As currently phrased, the "finally" part of the common steps in the installation section can be a bit confusing, as it may look like it tells you that the first step is to do something in a virtual environment, which is confusing if you're not using one, as instead of reading further and noticing it has instructions for other environments, you want instead go back to read earlier steps to figure out if you missed something that asked you to set up one.

Rephrase to make it more obvious that this subsection is a list of alternatives rather than a series of steps.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -199,35 +199,35 @@ If you think you might want to commit back to Bikeshed,
 instead download it over SSH.
 I won’t explain how to do that here.)
 
-Finally, run:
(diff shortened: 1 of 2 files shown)


tabatkins:
Fixed conflicts and committed in https://github.com/tabatkins/bikeshed/commit/f7860c06eaec18afc415bf3f4a7b90aee58ca680, thanks!

PR #1473, opened 2019-05-16
add .DS_Store to .gitignore

marcoscaceres:
Because life on MacOS is a joy sometimes...


domenic:
For issues specific to your system, I suggest using global gitignore, not changing every project you work on: http://egorsmirnov.me/2015/05/04/global-gitignore-file.html

marcoscaceres:
Yeah good point. Just got this Mac recently and forgot about that.

PR #1481, opened 2019-05-23
fix(metadata): use shortname in This Version link

### LINKED ISSUE #1480: "This version:" link generated incorrectly
The short name included in the "This version" link is being lower-cased, which causes W3C Specberus/PubRules to reject it. 

For example, running`bikeshed spec --md-status=WD` for the [File API](https://github.com/w3c/FileAPI) spec yields (note *fileapi*):

> https://www.w3.org/TR/2019/WD-fileapi-20190523/

But expected (FileAPI): 

> https://www.w3.org/TR/2019/WD-FileAPI-20190523/

marcoscaceres:
fixes #1480 

@tabatkins, don't know the code very well, so thought I'd try this first ☺️


marcoscaceres:
This is no good... needs to be vshortname... but vshortname can't have shorname lowercased.

PR #1482, opened 2019-05-23
Fix tests

marcoscaceres:
Fix build ... and apparently I can't spell https 🤦‍♂️

PR #1483, opened 2019-05-23
fix(metadata): don't lcase shortname in This Version link

marcoscaceres:
This includes https://github.com/tabatkins/bikeshed/pull/1482 otherwise tests won't pass... will rebase after https://github.com/tabatkins/bikeshed/pull/1482 goes in.


tabatkins:
Fixed more properly, since I *do* need to lowercase the shortname for linking purposes, by adding a displayVshortname and using that for the `[VSHORTNAME]` macro, which is used to construct the "This Version" link, in 6dcfb71

marcoscaceres:
Thanks for the speedy turn around on this @tabatkins.

PR #1500, opened 2019-07-29
Match more GitHub repository URL formats

chasephillips:
A GitHub repo can be `git clone`d with more than one URL format.
For example:
  git@github.XYZ:foo/bar.git
  https://github.XYZ/foo/bar.git
  https://github.XYZ/foo/bar
  https://github.XYZ/foo/bar/

.. all point to the same GitHub repo.  After the repo is cloned,
`git remote -v` will print the URL that was used to clone.  Bikeshed
previously only understood the first 2 of those 4 URL forms.

This PR adds the last 2 of those 4 URL forms to |getSpecRepository|
so that the GitHub repository is correctly computed in those cases.


chasephillips:
FYI @inexorabletash this PR fixes the issue we were looking into Friday.

PR #1508, opened 2019-08-19
Convert widlparser to LF newlines (was mixed CRLF/LF)

foolip:
constructs.py, productions.py and tokenizer.py had mixed CRLF and LF
line endings, making editing those files challenging.

Only a few other files were CRLF, so convert all to LF.


foolip:
This would conflict with https://github.com/tabatkins/bikeshed/pull/1509 and I'd recommend merging that first. This change can be reproduced using `dos2unix` on all files in the widlparser directory.

tabatkins:
As with the other patch, please make this change upstream, in plinss/widlparser. I won't take local changes unless that project dies.

foolip:
Trying to port this to https://github.com/plinss/widlparser/ I found that upstream doesn't fix CRLF and LF within any single file. @tabatkins it looks like what you have isn't a byte-for-byte copy, perhaps something to try forcing the next time you sync.

tabatkins:
I... literally just delete the whole folder and re-clone (then delete the `.git`/etc folders). Maybe git is doing some linefeed magic to "match platform" when I clone? In only some files?????

foolip:
Are you on Windows? If you use some autocrlf setting, that might explain stuff like this.

Issue #1507, opened 2019-08-19
Allow `async` as argument name


foolip:
https://github.com/tabatkins/bikeshed/commit/450e0499b28bb4c684052ad936ad1e7d36c4eac5 added `async` to `SymbolIdents` in tokenizer.py, causing it to be treated as a symbol rather than an identifier.

This caused a problem in https://github.com/whatwg/xhr/issues/251:
> FATAL ERROR: IDL SYNTAX ERROR LINE: 42 - skipped: "void open(ByteString method, USVString url, boolean async, optional USVString? username = null, optional USVString? password = null)"

https://github.com/heycam/webidl/pull/769 fixed this in the spec, this issue should probably be resolved by matching that, but I've framed it around the regression only as that's blocking https://github.com/whatwg/xhr/issues/252.

foolip:
Dupe of https://github.com/plinss/widlparser/issues/42.

PR #1509, opened 2019-08-19
Allow `async` as argument name

foolip:
Fixes https://github.com/tabatkins/bikeshed/issues/1507

### DIFF
--- a/bikeshed/widlparser/widlparser/productions.py
+++ b/bikeshed/widlparser/widlparser/productions.py
@@ -1059,7 +1059,7 @@ class Default(Production):   # "=" ConstValue | "=" string | "=" "[" "]" | "=" "
 
 
 class ArgumentName(Production):   # identifier | ArgumentNameKeyword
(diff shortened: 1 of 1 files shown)


foolip:
Note that this *only* fixes the problem for argument names, there's no `AttributeNameKeyword` variable in Bikeshed and I didn't spend time looking what should change instead.

foolip:
@tabatkins I see that Travis isn't running and thus can't confirm if this fixed the problem. Maybe you haven't enabled "Build pushed pull requests" in the Travis config?

domenic:
Note that this should ideally be updated at the source, i.e. https://github.com/plinss/widlparser/. See https://github.com/plinss/widlparser/issues/42.

tabatkins:
Yeah, I'm not taking this as a local patch unless it's clear widlparser isn't going to update. My update process for that folder is "delete and reclone from source".

foolip:
Oh, didn't realize that this was a 3p dependency, thought that commit 450e0499b28bb4c684052ad936ad1e7d36c4eac5 just had an overly generic commit message :)

Issue #1513, opened 2019-08-29
Update for constructor operation support


Ms2ger:
Probably just needs https://github.com/plinss/widlparser/issues/44 to be imported.

plinss:
widlparser has been updated.

Note that extended attribute constructors are exposed by the 'constructors' property, but constructor operations are not, they're just regular methods. Both are exposed by the 'members' property but constructor operations have a 'method' idlType while the [Constructor] and [NamedConstructor] attributes have a 'constructor' idlType.

Let me know if this is going to be an issue for bikeshed, but it was necessary to differentiate them within widlparser.

foolip:
Sent https://github.com/tabatkins/bikeshed/pull/1514 if anyone has the power to review.

birtles:
It seems like this doesn't linkify `constructor`? (Unlike `[Constructor]` which was link-ified.) And it doesn't correctly linkify the arguments of the constructor either with constructor scope?

Is that right, or am I just doing it wrong?

foolip:
That's right, see https://github.com/tabatkins/bikeshed/issues/1519.

birtles:
Thank you!

Issue #1506, opened 2019-08-19
Require tests to pass before pushing to master


foolip:
https://github.com/tabatkins/bikeshed/commit/450e0499b28bb4c684052ad936ad1e7d36c4eac5 made changes that caused Travis to fail:
https://travis-ci.org/tabatkins/bikeshed/builds/569899644

```
✘ 401/404 tests passed.
Failed Tests:
* github/whatwg/xhr/review-drafts/2018-08.bs
* github/whatwg/xhr/review-drafts/2019-02.bs
* github/whatwg/xhr/xhr.bs
```

This is https://github.com/whatwg/xhr/issues/251.

I've confirmed that 404/404 tests pass locally on the parent commit, 220086d88511a9c99d7a1f9b5447db7e7b99e053.

The most straightforward way to avoid this would be to protect the master branch and have all changes go in as PRs. But that's an implementation detail, anything that works... works.

domenic:
Oh, yes, this would have been absolutely ideal. Then we would have noticed the issue and worked to fix it on the Web IDL spec/widlparser side, before it started causing individual spec build failures.

tabatkins:
That wasn't a bug in Bikeshed: per the original WebIDL change, XHR was indeed invalid. We just, subsequently, made another change to WebIDL allowing it. (Tho not in this instance, apparently; per <https://github.com/whatwg/xhr/issues/251> `async` as an argument name is still invalid in the spec.)

Note that XHR did *not* fail to build, it just didn't mark up that line in its IDL block. (Unless its build process just totally breaks on error? I dunno what XHR's build process is, but Bikeshed is intended to be resilient to errors like this.)

foolip:
https://resources.whatwg.org/build/deploy.sh indeed fails on warnings (`-F die-on=warning`) and that's why this became blocking for https://github.com/whatwg/xhr/pull/252. (The same script is run on PRs, it just skips the final bit of actually deploying.)

But from Bikeshed's PoV, what happened is that https://travis-ci.org/tabatkins/bikeshed/builds went from green to red with that change, as `bikeshed test` went from passing to failing. Noticing when that happens is always a good idea even if the resolution is just to rebase the tests if the changes are intended.

tabatkins:
Oh I noticed, and filed <https://github.com/whatwg/xhr/issues/251> an hour and a half after making the change. ^_^

foolip:
Oh, so you did :)

[...]

PR #1504, opened 2019-08-08
Automatic test update

autofoolip:
By https://github.com/foolip/bikeshed-tests

PR #1514, opened 2019-08-30
Update widlparser.

foolip:
Update widlparser.

A verbatim copy of https://github.com/plinss/widlparser/commit/51d4103a79e04b061d8d462e4e03e8fb0e174e2b.

Brings in these changes:
https://github.com/plinss/widlparser/compare/2d94815b578f337d205ca06d5dd9fe39f6339783...51d4103a79e04b061d8d462e4e03e8fb0e174e2b

The diff in bikeshed appears bigger because there are also changes to
line endings that didn't match widlparser:
https://github.com/tabatkins/bikeshed/pull/1508

.gitignore and .hgignore were also pre-existing in widlparser but just
not copied to Bikeshed before.

Fixes https://github.com/tabatkins/bikeshed/issues/1513.


foolip:
@plinss do you have write access to this repo, or does only @tabatkins?

Travis isn't run on PRs it seems, but `bikeshed test` fails locally:
```
✘ 401/404 tests passed.
Failed Tests:
* github/w3c/csswg-drafts/web-animations-2/Overview.bs
* github/w3c/web-animations/Overview.bs
* github/w3c/webappsec-uisecurity/index.bs
```

That's different from the failures on master locally:
```
✘ 401/404 tests passed.
Failed Tests:
* github/whatwg/xhr/review-drafts/2018-08.bs
* github/whatwg/xhr/review-drafts/2019-02.bs
* github/whatwg/xhr/xhr.bs
```

https://github.com/tabatkins/bikeshed/issues/1506

foolip:
I've bisected those new test failures to https://github.com/plinss/widlparser/commit/c1f30088df0d1cf460acdcbe01ff5e4bbcb2640c. Looks like there's some difference when there's whitespace in a method, like `void foo ();` instead of `void foo();`.

foolip:
I've filed https://github.com/plinss/widlparser/issues/47 for the regression, which also affects attribute names. I can't work out precisely what the problem is, so this is blocked :(

plinss:
Regression fixed, please update to the latest widlparser.

I don't have write access to this repo.

foolip:
Then @tabatkins will have to help. I'll update this next week before he's back at work.

foolip:
I've updated again and force pushed, but now there are other changes:
```
✘ 400/404 tests passed.
Failed Tests:
* idl001.bs
* idl003.bs
* github/w3c/IndexedDB/index.bs
* github/w3c/csswg-drafts/cssom-1/Overview.bs
```

The changes all seem related to underscores.

foolip:
This is now blocked on https://github.com/plinss/widlparser/issues/51 instead.

foolip:
Updated again to also roll in https://github.com/plinss/widlparser/pull/52, but the `bikeshed test` changes remain unresolved. Some of this is just updating IDs to not include underscores, and that seems harmless. I'll push the updated baselines to be able to comment here, but they should not be landed as-is.

tests/github/w3c/IndexedDB/index.html
```diff
@@ -3360,7 +3360,7 @@ <h3 class="heading settled" data-level="4.7" id="keyrange"><span class="secno">4
...
 
-  <a class="idl-code" data-link-type="interface" href="https://heycam.github.io/webidl/#idl-boolean" id="ref-for-idl-boolean①②"><c- b>boolean</c-></a> <a class="idl-code" data-link-type="method" href="#dom-idbkeyrange-_includes" id="ref-for-dom-idbkeyrange-_includes"><c- g>_includes</c-></a>(<c- b>any</c-> <dfn class="idl-code" data-dfn-for="IDBKeyRange/_includes(key)" data-dfn-type="argument" data-export id="dom-idbkeyrange-_includes-key-key"><code><c- g>key</c-></code><a class="self-link" href="#dom-idbkeyrange-_includes-key-key"></a></dfn>);
+  <a class="idl-code" data-link-type="interface" href="https://heycam.github.io/webidl/#idl-boolean" id="ref-for-idl-boolean①②"><c- b>boolean</c-></a> <a class="idl-code" data-link-type="method" href="#dom-idbkeyrange-_includes" id="ref-for-dom-idbkeyrange-_includes"><c- g>_includes</c-></a>(<c- b>any</c-> <dfn class="idl-code" data-dfn-for="IDBKeyRange/includes(key)" data-dfn-type="argument" data-export id="dom-idbkeyrange-includes-key-key"><code><c- g>key</c-></code><a class="self-link" href="#dom-idbkeyrange-includes-key-key"></a></dfn>);
```
foolip: The change removes the underscore in `data-dfn-for="IDBKeyRange/_includes(key)"`, `id="dom-idbkeyrange-_includes-key-key"` and `href="#dom-idbkeyrange-_includes-key-key"`. That by itself seems harmless.

tests/github/w3c/csswg-drafts/cssom-1/Overview.html
```diff
@@ -1898,7 +1898,7 @@ <h4 class="heading settled" data-level="6.6.1" id="the-cssstyledeclaration-inter
...
+  [<a class="idl-code" data-link-type="extended-attribute" href="https://html.spec.whatwg.org/multipage/custom-elements.html#cereactions" id="ref-for-cereactions④"><c- g>CEReactions</c-></a>] <c- b>attribute</c-> [<a class="idl-code" data-link-type="extended-attribute" href="https://heycam.github.io/webidl/#TreatNullAs" id="ref-for-TreatNullAs④"><c- g>TreatNullAs</c-></a>=<a class="n" data-link-type="idl-name"><c- n>EmptyString</c-></a>] <a class="n" data-link-type="idl-name" href="#cssomstring" id="ref-for-cssomstring②⑤"><c- n>CSSOMString</c-></a> <dfn class="idl-code" data-dfn-for="CSSStyleDeclaration" data-dfn-type="attribute" data-export data-lt="camel_cased_attribute" data-type="[TreatNullAs=EmptyString] CSSOMString" id="dom-cssstyledeclaration-camel_cased_attribute"><code><c- g>_camel_cased_attribute</c-></code><a class="self-link" href="#dom-cssstyledeclaration-camel_cased_attribute"></a></dfn>;
```
foolip: This adds `data-lt="camel_cased_attribute"` and removes some underscores, but that seems OK. The actual visible text still has the underscore.

tests/github/w3c/csswg-drafts/cssom-1/Overview.html
```diff
@@ -2577,7 +2577,7 @@ <h3 class="no-num no-ref heading settled" id="index-defined-here"><span class="c
    <li><a href="#concept-css-style-sheet-alternate-flag">alternate flag</a><span>, in §6.1</span>
    <li><a href="#dom-medialist-appendmedium">appendMedium(medium)</a><span>, in §4.4</span>
    <li><a href="#associated-css-style-sheet">associated CSS style sheet</a><span>, in §6.3.2</span>
-   <li><a href="#dom-cssstyledeclaration-_camel_cased_attribute">_camel_cased_attribute</a><span>, in §6.6.1</span>
+   <li><a href="#dom-cssstyledeclaration-camel_cased_attribute">camel_cased_attribute</a><span>, in §6.6.1</span>
```
foolip: This does change the visible text, but is probably fine?

tests/idl001.html
```diff
@@ -373,10 +373,10 @@ <h2 class="no-num no-ref heading settled" id="idl-index"><span class="content">I
     <li><a href="#ref-for-dom-foo-foo">Unnamed section</a>
    </ul>
   </aside>
-  <aside class="dfn-panel" data-for="dom-foo-_or">
-   <b><a href="#dom-foo-_or">#dom-foo-_or</a></b><b>Referenced in:</b>
+  <aside class="dfn-panel" data-for="dom-foo-or">
+   <b><a href="#dom-foo-or">#dom-foo-or</a></b><b>Referenced in:</b>
    <ul>
-    <li><a href="#ref-for-dom-foo-_or">Unnamed section</a> <a href="#ref-for-dom-foo-_or①">(2)</a>
+    <li><a href="#ref-for-dom-foo-or">Unnamed section</a>
```
foolip: This change I don't understand and looks like it could be a regression. @tabatkins @plinss?

Ms2ger: Linking with `{{or}}` still works, but `{{_or}}` doesn't. I think that's fine.

foolip: Yeah, that does seem just fine.

plinss:
the current version of widlparser strips leading underscores from the names of all identifiers now, likely to be more of these cases.

foolip:
As it turns out only two real-world specs use underscores in a way that changes the output with the new widlparser changes.

I've merged in https://github.com/tabatkins/bikeshed/pull/1504 and run `bikeshed test` locally to see if any newer spec changes would break, and there are additional differences:
```
✘ 406/410 tests passed.
Failed Tests:
* github/heycam/webidl/index.bs
* github/whatwg/xhr/review-drafts/2018-08.bs
* github/whatwg/xhr/review-drafts/2019-02.bs
* github/whatwg/xhr/xhr.bs
```

However, it looks to me like the differences are intended ones. @tabatkins can you merge https://github.com/tabatkins/bikeshed/pull/1504 first?

foolip:
Uh, so changes were made for https://github.com/plinss/widlparser/issues/51 which should also be rolled in, but running or rebasing the tests is broken now: https://github.com/tabatkins/bikeshed/issues/1517

@tabatkins can you take it from here?

tabatkins:
Is the "strip underscores from all identifiers" actually what WebIDL says to do now? I haven't merged the update yet because I'm not sure about this.

tabatkins:
Yeah as far as I can tell that's not what's changed? You're still not supposed to be able to make, say, a method named `async` from what I can tell.

Or is this all fine and just a linking issue, so I just need to make sure Bikeshed is happy with stuff (auto-stripping underscores from linking texts, I assume, so it remains just something in the actual idl text)? I'm not sure.

plinss:
AFAICT WebIDL does say to strip the leading underscore from all identifiers as a form of escaping, see: https://heycam.github.io/webidl/#ref-for-dfn-identifier 

Note that I do find the prose here to be a bit confusing and I suppose it could be interpreted to mean "only strip the leading underscore when the identifier is a reserved word", but I don't really see text to support that. Clarifications are welcome.

widlparser does preserve the leading underscores in the IDL itself, it just removes them from names returned in the API. So you *can* have a method named 'async', e.g.:

    interface Foo {
        long _async();
    };

widlparser will now return 'async' as the name of that method, but the IDL will still have `_async()`. (not clear to me how you'd actually call that method from JS, is `foo.async()` valid?)

domenic:
`foo.async()` is valid; JS has no restrictions on method names. (Only variable names.)

Ms2ger:
> Note that I do find the prose here to be a bit confusing and I suppose it could be interpreted to mean "only strip the leading underscore when the identifier is a reserved word", but I don't really see text to support that. Clarifications are welcome.

In the example below, there's:

> Note that while the second attribute on the TextField interface need not have been escaped with an underscore (because "value" is not a keyword in the IDL grammar), it is still unescaped to obtain the attribute’s identifier.

Still, that section could use some love.

plinss:
Fair enough, so widlparser's behavior is correct then. (I shouldn't read specs on less than 4 hours of sleep.)

tabatkins:
All right, I updated manually. (For future reference, I remove the .git* and .hg* files on purpose, as I'm not interested in retaining the version history of my deps.)

PR #1554, opened 2019-11-26
Use https links in W3C copyright statements

plehegar:
This aligns with https://www.w3.org/pubrules/doc/rules/?profile=WD#copyright


plehegar:
cc @akshayku @wseltzer

PR #1555, opened 2019-12-01
imply plain printMode on dumb terminals

andreastt:
Dumb terminals only recognise a limited subset of escape sequences
and are identified by TERM=dumb.

With this change, bikeshed will imply the plain printMode (equal
to NO_COLOR=1) for dumb terminals.


tabatkins:
Just to be sure - is "TERM" present in all environments? If not this'll crash.

andreastt:
Of course we should check for that, thanks!

PR #1557, opened 2019-12-03
Add boilerplate for CSS Recommendations

frivoal:
Based on the status section of https://www.w3.org/TR/2019/REC-css-contain-1-20191121/

### DIFF
--- a/bikeshed/boilerplate/csswg/status-REC.include
+++ b/bikeshed/boilerplate/csswg/status-REC.include
@@ -1,0 +1,53 @@
+<p>
+	<em>This section describes the status of this document
+	at the time of its publication.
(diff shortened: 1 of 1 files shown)


frivoal:
Actually, never mind. Reason best explained by @svgeesus:

> We *never* make the Rec from the ED [1]. 
>
> Instead, we make the Rec from the Proposed Rec, changing the status and styling to suit. (And if actually needed, we apply any editorial changes which were agreed with the AC, and which we talk about in the Director's Decision).
>
> [1] Because we don't add in whatever changes might have been made to the ED in the couple of months since the Proposed Rec was generated, which the AC has not seen.

PR #1562, opened 2019-12-06
Fix a typo in Shepherd URL in test

foolip:


foolip:
@tabatkins I guess this URL doesn't matter to the test? Anyway I found it in code search and failed to follow it :)

tabatkins:
That code only gets run if you call that file from the command line, so no, it doesn't matter to the tests. ^_^

PR #1564, opened 2019-12-08
Add new feature for injecting MDN panels

sideshowbarker:
This change adds a new feature for automatically including MDN panels (annotations) in spec output. The feature is controlled by the "Include Mdn Panels" boolish.

Sample output for several specs:

* https://domspec.herokuapp.com/
* https://fetchspec.herokuapp.com/
* https://streamsspec.herokuapp.com/
* https://urlspec.herokuapp.com/
* https://xhrspec.herokuapp.com
* https://encodingspec.herokuapp.com/
* https://webidlspec.herokuapp.com/
* https://serviceworkerspec.herokuapp.com/docs/
cc @annevk @domenic

### DIFF
--- a/bikeshed/Spec.py
+++ b/bikeshed/Spec.py
@@ -11,6 +11,7 @@ from datetime import datetime
 from . import biblio
 from . import boilerplate
 from . import caniuse
+from . import mdnspeclinks
 from . import config
 from . import constants
 from . import datablocks
@@ -73,6 +74,7 @@ class Spec(object):
         self.typeExpansions = {}
         self.macros = defaultdict(lambda x: "???")
         self.canIUse = {}
+        self.mdnSpecLinks = {}
         self.widl = idl.getParser()
         self.testSuites = json.loads(self.dataFile.fetch("test-suites.json", str=True))
         self.languages = json.loads(self.dataFile.fetch("languages.json", str=True))
@@ -200,6 +202,7 @@ class Spec(object):
         biblio.dedupBiblioReferences(self)
         verifyUsageOfAllLocalBiblios(self)
         caniuse.addCanIUsePanels(self)
+        mdnspeclinks.addMdnPanels(self)
         boilerplate.addIndexSection(self)
         boilerplate.addExplicitIndexes(self)
         boilerplate.addStyles(self)
--- a/bikeshed/cli.py
+++ b/bikeshed/cli.py
@@ -95,6 +95,7 @@ def main():
     updateParser.add_argument("--backrefs", action="store_true", help="Download link backref data.")
     updateParser.add_argument("--biblio", action="store_true", help="Download biblio data.")
     updateParser.add_argument("--caniuse", action="store_true", help="Download Can I Use... data.")
+    updateParser.add_argument("--mdn", action="store_true", help="Download MDN Spec Links... data.")
     updateParser.add_argument("--link-defaults", dest="linkDefaults", action="store_true", help="Download link default data.")
     updateParser.add_argument("--test-suites", dest="testSuites", action="store_true", help="Download test suite data.")
     updateParser.add_argument("--languages", dest="languages", action="store_true", help="Download language/translation data.")
@@ -234,7 +235,7 @@ def main():
 
 
 def handleUpdate(options, extras):
-    update.update(anchors=options.anchors, backrefs=options.backrefs, biblio=options.biblio, caniuse=options.caniuse, linkDefaults=options.linkDefaults, testSuites=options.testSuites, languages=options.languages, wpt=options.wpt, dryRun=constants.dryRun, force=options.force)
+    update.update(anchors=options.anchors, backrefs=options.backrefs, biblio=options.biblio, caniuse=options.caniuse, mdn=options.mdn, linkDefaults=options.linkDefaults, testSuites=options.testSuites, languages=options.languages, wpt=options.wpt, dryRun=constants.dryRun, force=options.force)
 
 
 def handleSpec(options, extras):
--- a/bikeshed/mdnspeclinks.py
+++ b/bikeshed/mdnspeclinks.py
@@ -84,16 +84,16 @@ def addMdnPanels(doc):
     #       https://github.com/whatwg/whatwg.org/pull/267 resolved
     doc.extraStyles["style-mdn-anno"] = '''
         @media (max-width: 767px) { .mdn-anno { opacity: .1 } }
-        .mdn-anno { font: 1em sans-serif; padding: 0.3em; position: absolute; z-index: 8; top: auto; right: 0.3em; background: #EEE; color: black; box-shadow: 0 0 3px #999; overflow: hidden; border-collapse: initial; border-spacing: initial; min-width: 9em; max-width: min-content; overflow-wrap: normal; word-wrap: normal; hyphens: none}
+        .mdn-anno { font: 1em sans-serif; padding: 0.3em; position: absolute; z-index: 8; top: auto; right: 0.3em; background: #EEE; color: black; box-shadow: 0 0 3px #999; overflow: hidden; border-collapse: initial; border-spacing: initial; min-width: 9em; white-space: nowrap; overflow-wrap: normal; word-wrap: normal; hyphens: none}
         .mdn-anno:not(.wrapped) { opacity: 1}
         .mdn-anno:hover { z-index: 9 }
         .mdn-anno.wrapped { min-width: 0 }
         .mdn-anno.wrapped > :not(button) { display: none; }
-        .mdn-anno > .mdn-anno-btn { cursor: pointer; border: none; color: #000; background: transparent; margin: -8px; outline: none; }
+        .mdn-anno > .mdn-anno-btn { cursor: pointer; border: none; color: #000; background: transparent; margin: -8px; float: right; padding: 10px 8px 8px 8px; outline: none; }
         .mdn-anno > .mdn-anno-btn > .less-than-two-engines-flag { color: red; padding-right: 2px; }
         .mdn-anno > .mdn-anno-btn > .all-engines-flag { color: green; padding-right: 2px; }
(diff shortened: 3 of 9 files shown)


sideshowbarker:
I noticed something broken.

This works:

https://serviceworkerspec.herokuapp.com/docs/#service-worker-global-scope-install-event

…in that, I can see an MDN annotation for the `install` event, in the right place.

But if the spec ToC UI, I then push the arrow to collapse the ToC sidebar, the MDN anno disappears.

If I re-expand the ToC sidebar, the MDN anno reappears in the expected place.

annevk:
For WHATWG specifications, I don't think we want Bikeshed to deliver the style sheets and script, but rather have those as shared external resources across all our specifications. So ideally those become options.

sideshowbarker:
> For WHATWG specifications, I don't think we want Bikeshed to deliver the style sheets and script, but rather have those as shared external resources across all our specifications. So ideally those become options.

Since this code’s already loading the browser icons from https://resources.whatwg.org, I wonder if we could make the (final) code also always load the script and stylesheet for this feature from https://resources.whatwg.org, regardless?

sideshowbarker:
> But if the spec ToC UI, I then push the arrow to collapse the ToC sidebar, the MDN anno disappears.

Fixed.

annevk:
I suspect the W3C would not want to fetch third-party resources, much like the WHATWG also tries to avoid third-party resources.

sideshowbarker:
> I suspect the W3C would not want to fetch third-party resources, much like the WHATWG also tries to avoid third-party resources.

I don’t know what the policy is, but the current code in Bikeshed for the existing Can I Use panels feature uses the https://resources.whatwg.org/ browser icons. So any specs published at the W3C which have Can I Use panels are already fetching those.

But that said, I guess fetching third-party scripts would be a different issue, as far as policy

domenic:
Probably we want the design similar to [`Use Dfn Panels`](https://tabatkins.github.io/bikeshed/#metadata-use-dfn-panels):

> It automatically adds some JS and CSS to your page for the panels; you can suppress this with `Boilerplate: script-dfn-panel no, style-dfn-panel no` and supply your own if desired.

tabatkins:
Yup, that's the expected way. Add the script/styles to the `Spec.extraStyles` or `Spec.extraScripts` dicts, with some appropriate key, when the feature is turned on and has at least one thing to script/style at. That'll automatically add them to the document by default, but give the free suppression behavior via `Boilerplate`.

tabatkins:
All right, fixed up and merged.

Final question remains: why are we keeping in the `data-mdn-for` attributes for a Slim Build Artifact? I think those wouldn't have the annos turned on in the first place.

tabatkins:
Okay, just tried the feature out on the DOM spec, and it mostly looks fine, except that when expanded the panel fills the entire screen-width. Presumably something's up with the CSS, but I haven't tried to diagnose it yet.

----

Also, looking into the structure of the MDN data a little more, I don't think that the top-level keys actually map to IDs, at least not reliably.

For example, check out <https://w3c.github.io/mdn-spec-links/accelerometer.json> and search for "accelerometer-y". It has items in its "features" array - one is the `Accelerometer.y` attribute, but the second is `LinearAccelerationSensor.y`, a different feature (a subclass of `Accelerometer`) with a different ID.  As currently set up, `Accelerometer.y` would get the panel, with two sets of data in it.

However, I have no idea how to derive the correct ID here; the data simply doesn't seem to be present. :/

sideshowbarker:
> Okay, just tried the feature out on the DOM spec, and it mostly looks fine, except that when expanded the panel fills the entire screen-width. Presumably something's up with the CSS, but I haven't tried to diagnose it yet.


Lack of `max-width`, I think. I had added `max-width: min-content` previously, but I removed it in https://github.com/tabatkins/bikeshed/pull/1564/commits/dc0d6f2624973db36536eb553d6b0c0e675ebac8 because I couldn’t remember why I had added it…

I also added `white-space: nowrap` in that change. I’m not sure now why, when I added `white-space: nowrap`, I deleted `max-width: min-content` at the same time.

The problem I am trying to solve is that I don’t want any hyphenated MDN feature names/links to be break/wrap at the hyphens — so I was trying to avoid constraining the width of the panels to the point that those MDN feature names/links end up getting wrapped at the hyphens.

So, I reckon there’s some best-practice way to do that in CSS that I just haven’t hit on yet.

In the meantime, I’ll try re-adding `max-width: min-content` and see if that fixes it.

sideshowbarker:
> Also, looking into the structure of the MDN data a little more, I don't think that the top-level keys actually map to IDs, at least not reliably.

Right. It’s only as reliable as the data in the Specifications sections of MDN articles.

The upstream BCD data — from the https://github.com/mdn/browser-compat-data repo, which is a different thing than the MDN articles — doesn’t yet have mappings for features to particular spec URLs. (Well, it does only for the case of JavaScript features, because that’s the only subset for which the BCD maintainers have so have agreed to add it). Instead, the spec URLs are only in MDN articles, and to get them, I have a setup that regularly scrapes MDN to extract them.

> For example, check out https://w3c.github.io/mdn-spec-links/accelerometer.json and search for "accelerometer-y". It has items in its "features" array - one is the `Accelerometer.y` attribute, but the second is `LinearAccelerationSensor.y`, a different feature (a subclass of `Accelerometer`) with a different ID. As currently set up, `Accelerometer.y` would get the panel, with two sets of data in it.
> 
> However, I have no idea how to derive the correct ID here; the data simply doesn't seem to be present. :/

So to be clear, there’s no magic I’m trying to do anywhere to derive the correct IDs from feature names. Instead, the mappings in https://w3c.github.io/mdn-spec-links are just generated from scraping MDN articles and pulling out the spec URLs, with the fragments.

So it’s kind of a garbage-in-garbage-out case. But that said, the end result is mostly non-garbage — because the input is mostly non-garbage. However, I now and then run into exceptions.

In the particular case of `LinearAccelerationSensor.y`, I think the problem described is just due to the https://developer.mozilla.org/en-US/docs/Web/API/LinearAccelerationSensor/y#Specifications having a reference to the wrong spec fragment ID — it wrongly references https://www.w3.org/TR/accelerometer/#accelerometer-y when it should be referencing https://www.w3.org/TR/accelerometer/#linearaccelerationsensor-y. (Well, and probably it also shouldn’t be referencing the TR-space draft to begin with, but that’s a different problem.)

So anyway, that’s the kind of garbage-in-garbage-out case I meant. I can’t really programatically detect errors in the MDN articles such as that one. They just need to be fixed when we find them — but updating the MDN articles to have the right IDs, and then regenerating the https://w3c.github.io/mdn-spec-links JSON files.

Actually part of the value getting the MDN annotations added to the specs is to expose cases like that so we can fix them.

sideshowbarker:
> Final question remains: why are we keeping in the `data-mdn-for` attributes for a Slim Build Artifact? I think those wouldn't have the annos turned on in the first place.

I think the reason I added the code to ensure the `data-mdn-for` attributes get kept is that, without that, I otherwise found that the attributes got dropped when we actually needed them to not be dropped. I mean I found that during testing. I can’t remember now if I observed that only in some cases — with certain specs — or in all cases. I tested with a ton of different specs…

But maybe the case where I ran into that problem was when trying to get the MDN annotations into the output of the “review draft” of the DOM spec. Maybe that and other WHATWG “review draft” builds are configured to use Slim Build? (I need read up on Slim Build; until the mention in this issue, I had never heard of it…)

Anyway, for now what I’ll do is, I’ll remove that code locally and see if things work as expected without it — including with the review-draft build of the DOM spec.

domenic:
I am pretty sure the slim build stuff was built specifically for review drafts. But that said, it would probably make sense for slim build to exclude MDN stuff by default, with some way to explicitly turn them back on in the WHATWG configuration for RDs.

sideshowbarker:
> it would probably make sense for slim build to exclude MDN stuff by default, with some way to explicitly turn them back on in the WHATWG configuration for RDs.

Yeah, I agree. If somebody’s explicitly opting-in to a “slim build”, then they should get a slim build — and having the MDN stuff in there doesn’t make it slim.

So we should flip the logic around such that the slim-build option does cause them to be omitted, and then … add another option that overrides that?

The thing is, I think maybe we don’t even want the MDN annos in all WHATWG RDs. We really only want them in the very specific case that we’re generating a W3C-ready review draft. And per the MoU, the only specs we need that for are DOM and HTML (which of course this discussion isn’t relevant to, since HTML isn’t using Bikeshed).

domenic:
Good point.

I'm not sure of exactly how slim build works, but I'm kind of hoping it's a macro of sorts that turns off a bunch of stuff, and then if you explicitly turn something back on, the latter will win. In which case we could just turn it back on for DOM RDs in particular.

sideshowbarker:
> I'm not sure of exactly how slim build works, but I'm kind of hoping it's a macro of sorts that turns off a bunch of stuff, and then if you explicitly turn something back on, the latter will win.

Here’s the relevant code:

https://github.com/tabatkins/bikeshed/blob/d335102afd6b0e834ff899862c9286f712927622/bikeshed/unsortedJunk.py#L1153-L1157

So `doc.md.slimBuildArtifact` isn’t really handled in the code as a macro kind of thing.

And even if, say, we changed the condition there to:

```python
if doc.md.slimBuildArtifact and not doc.md.includeMdnPanels
```

…then that’d cause *all* `data-` attributes to be preserved — which is not what we need or want.

This makes me remember now that I had actually considered using something other than a `data-` attribute to associate the annotation with the element it’s annotating. But I don’t see any other appropriate global attribute that could be used — and this pretty much the kind of thing which we created `data-` attributes for to begin with.

tabatkins:
Yeah, the point of `Slim Build Artifact` was to remove as much as possible while still retaining a functional spec.

However, it was *added* for the purpose of generating RDs. If generating RDs actually requires something else, I'm fine with just swapping the functionality over to a more RD-focused name and purpose, rather than pretending this is a generic feature.

sideshowbarker:
> However, it was _added_ for the purpose of generating RDs. If generating RDs actually requires something else, I'm fine with just swapping the functionality over to a more RD-focused name and purpose, rather than pretending this is a generic feature.

Yeah unless we’re aware of `Slim Build Artifact` being used by other people for something, then we just naming it with `Review Draft` — or even `WHATWG Review Draft` — would help to make it a bit more obvious what it’s actually being used for.

PR #1578, opened 2020-01-07
WHATWG: add Tracking Vector Image to defaults

annevk:


annevk:
This depends on merging https://github.com/whatwg/whatwg.org/pull/64, but until we start using this that shouldn't be a problem.

annevk:
It's not clear to me how the test failures can possibly relate to this change.

annevk:
(This depends a bit on https://github.com/tabatkins/bikeshed/pull/1579 in that we don't want "potential" in our alt/title and would rather have a distinct alt with a space at the end.)

tabatkins:
The test failures aren't related, there's something weird between my local and test environments that I haven't nailed down yet.

PR #1579, opened 2020-01-07
Add Tracking Vector Text metadata

annevk:
And align with defaults used in https://github.com/whatwg/wattsi/pull/41.


annevk:
I haven't tried running this code. If CI doesn't take care of that I can do some more work.

tabatkins:
I don't quite understand what you're trying to achieve here - it looks like your TV Text is serving an identical purpose to the existing TV Title?  The only difference seems to be that if Bikeshed generates the SVG, it uses one for `title` and the other for `aria-label`, but to the best of my knowledge a `title` element in an SVG suffices as an assistive label already.

Is a space after alt text really something that authors need to worry about? That seems... worrying if true, since I doubt practically anyone in the world does that.

annevk:
I'm trying to reproduce what HTML is using for `` today (while accounting for the changes we want to make with respect to naming and seriousness).

Another thing I noticed is that by putting the image first, the text flow is a bit different, so perhaps the `alt` should be "This is a tracking vector: " now.

annevk:
And as that doesn't make sense as the title, I'd still like to slightly distinguish between those, but we could make it clear that they are to be almost identical.

annevk:
Any new thoughts @tabatkins?

annevk:
This got adopted and landed in master via 2ccc98e0ac694b1c243d3bf6d65b7678bc95a6c0 it seems. Yay! Time to start building stuff on top.

PR #1585, opened 2020-01-14
Automatic test update

autofoolip:
By https://github.com/foolip/bikeshed-tests


foolip:
@tabatkins no test update in a long time, would it be a good idea to land one to not risk regressing things in current specs? It looks like bikeshed-tests is no longer successfully updating this PR, though.

tabatkins:
Yes, I super need to land an update for this. :(

annevk:
Not sure why, but I decided to try rebase this. If it works I'll land this.

annevk:
It seems it would still be good for someone to fix https://github.com/tabatkins/bikeshed-tests, but not touching that today.

PR #1587, opened 2020-01-14
Wrap tracking vector image in link

annevk:
This requires https://github.com/tabatkins/bikeshed/pull/1586 and https://github.com/whatwg/infra/pull/115 to land first and can then be rebased to remove the first commit (that should not need reviewing here).


annevk:
Dependencies have landed. `bikeshed test --manual-only` passes.

annevk:
@tabatkins I'm not sure if that would work, but `<a lt="tracking vector">` might be an alternative here. Would you prefer that?

Issue #964, opened 2017-03-28
Add fingerprinting watermark when requested


tabatkins:
<https://github.com/whatwg/infra/issues/20> suggests Bikeshed auto-handling fingerprinting watermarks.  <https://github.com/whatwg/whatwg.org/pull/64> is an SVG of the fingerprint icon.  Example usage is at <https://html.spec.whatwg.org/multipage/webappapis.html#a-plausible-language>.

Tentative plan is:

1. Look for elements with `class=tracking-vector`.
2. Append a link-image with the fingerprint, pointing to some Infra explanation of the concept.
3. If at least one shows up, trigger Bikeshed-specific styling to put it in the margin. Same as other styles, you can turn it off if you have your own.
4. If at least two show up, put the SVG as a `defs` instead, and make the instances just have a `use`.
5. Add a "Fingerprint Image" metadata, taking a url, so you can point it to a specific image, instead of the Bikeshed-generated SVG.

zcorpan:
When using the image inline you need to add these attributes to the `<svg>`
```
width="46" height="64" role="img" aria-label="(This is a fingerprinting vector.)"
```

zcorpan:
Ping.

annevk:
@tabatkins ping?

annevk:
So FWIW, I think we want to slightly change this based on feedback. It's useful to have both "fingerprinting vector" and "tracking vector". The latter being a superset for some cases that are not covered by fingerprinting alone. See https://github.com/whatwg/infra/pull/115#issuecomment-353135658.

marcoscaceres:
Could this be a general "base.css" thing? https://github.com/w3c/tr-design/issues/162

annevk:
That could work for W3C documents and should probably be tracked separately.

Also, I don't think we need the complexity for now from my comment above. Doing "tracking vector" alone should be sufficient for the foreseeable future and encompasses both.

annevk:
@tabatkins ping and thank you!

tabatkins:
All right, phew, I took two years to do something that was only thirty minutes of actual work. As of f5a6d58a you can now activate the tracking-vector icon automatically.

This isn't documented yet, because I haven't adjusted the default Bikeshed stylesheets to accommodate it yet, but:

* by default, anything with `class=tracking-vector` will get a tracking-vector icon added as its first child (ideal for positioning into the margin with a `position:absolute; left: XXXpx;`)
[...]

PR #1590, opened 2020-01-15
Tracking Vector documentation

annevk:
This needs #1587 to land first and then I'll regenerate. It would also be better if Infra's tracking vector definition was indexed so no warnings are generated.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2040,7 +2040,13 @@ Tracking Vectors {#tracking-vectors}
 A <a>tracking vector</a> can be annotated with the `tracking-vector` attribute on the enclosing element.
 This will result in a link containing an image being prepended to the element.
 
-<p tracking-vector>This is an example of an element annotated with that attribute.
+<div class=example id=example-tracking-vector>
+  <xmp highlight=html>
+    <p tracking-vector>This is an example of an element annotated with that attribute.
+  </xmp>
+
+  <p tracking-vector>This is an example of an element annotated with that attribute.
+</div>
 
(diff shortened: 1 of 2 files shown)


annevk:
I guess we can also close #964 as part of landing this.

docs/index.bs
```diff
@@ -2027,6 +2034,18 @@ and so will work in include-raw substitutions,
 but that will change in the future.
 
 
+Tracking Vectors {#tracking-vectors}
+----------------
+
+A <a>tracking vector</a> can be annotated with the `tracking-vector` attribute on the enclosing element.
+This will result in a link containing an image being prepended to the element.
+
+<p tracking-vector>This is an example of an element annotated with that attribute.
```
tabatkins: Before this, put the exact same code in a `xmp highlight=html>` block

annevk:
Done. You don't use squash and merge by the way?

tabatkins:
I prefer linear history when I can, and for some reason they don't offer a squash-and-rebase option

annevk:
That's what squash and merge does.

annevk:
(At the WHATWG we also really like linear history.)

tabatkins:
> That's what squash and merge does.

Wait, really?  Then why do they call it "squash and merge"???

/me goes to look at documentation

tabatkins:
Well gosh dang it, indeed it does. It specifically enforces a fast-forward commit, which effectively rebases.

Okay, I'm changing the settings then, dang.

PR #1591, opened 2020-01-15
Optimize Dockerfile

emlun:
- Cache spec-data/ between builds, to greatly reduce rebuild time
- Remove unused files from the image

This is a big help when developing Bikeshed. :slightly_smiling_face:


emlun:
Wow, that was _really_ quick! :open_mouth:

tabatkins:
You caught me just as I was checking my email, and I trust that you did the right thing (since I don't understand dockerfiles myself) ^_^

emlun:
Hah, I suspected something like that. Thank you!

PR #1594, opened 2020-01-17
Fix rendering of default values [] and {} for dictionary members

### LINKED ISSUE #1592: DictionaryMember default [] is rendered as None
For example, see https://www.w3.org/TR/2019/WD-webauthn-2-20191126/#dom-publickeycredentialcreationoptions-excludecredentials, where this markup:

```
    sequence<PublicKeyCredentialDescriptor>      excludeCredentials = [];
```

is rendered as:

>***excludeCredentials*, of type sequence\<PublicKeyCredentialDescriptor\>, defaulting to `None`**

I'm new to the fine details of WebIDL and Bikeshed, but if I understand the specs correctly, `[]` is a valid default value and should probably be rendered as simply "[]". In that case, the root cause of the issue seems to be here:

https://github.com/tabatkins/bikeshed/blob/2a881f0ee698433d75c6a1555514334adb37d8c7/bikeshed/idl.py#L158-L160

which does not account for that `construct.default.value` will be `None` if the default value is `[]` or `{}`, in which case the object will have `._open` and `._close` set instead of `.value`.

Changing it to this seems to fix the issue, but it doesn't seem like the right solution:

```
                value = escapeAttr("{0}".format(construct.default.value if construct.default.value else (construct.default._open.symbol + construct.default._close.symbol)))
```

I'd be happy to contribute a fix if I can get a confirmation that my understanding seems correct, and perhaps some pointers to what might be a better solution. :slightly_smiling_face:

emlun:
This aims to fix #1592.

As of yet (commit 5f2dc38), this only adds a couple of tests.


tabatkins:
Ah, don't add tests to widlparser, that's a separate project (https://github.com/plinss/widlparser)

emlun:
Oh, ok. Thanks!

emlun:
Latest push includes a working fix which seems fine enough given the current widlparser API, but it might make sense to add a new method to `Default` in widlparser for this. See also plinss/widlparser#54. Expect another force-push once that's been settled.

tabatkins:
All right, I'm merging for now. The upstream fix in widlparser is in the Python 3 port, which I'll be integrating as I finish Bikeshed's *own* python 3 port next week, so just accepting your quick/hacky fix for now works just fine.

Thanks so much for noticing this and tracking down a fix!

emlun:
Great, thanks!

PR #1598, opened 2020-01-24
Allow linking to lower level in delta specs.

fantasai:


frivoal:
Note: the merge conflict indicated by github is a trivially resolved one.

bikeshed/refs/ReferenceManager.py
```diff
@@ -519,6 +522,20 @@ def getRef(
...
+                linkType=linkType,
+                spec=self.shortname,
+                status=status,
+                statusHint=statusHint,
+                linkFor=linkFor,
+                linkForHint=linkForHint,
+                explicitFor=explicitFor,
+                export=False,
+                ignoreObsoletes=True,
+                latestOnly=False)
```
nschonni, line 537:
```suggestion
                latestOnly=False,
            )
```

bikeshed/refs/ReferenceManager.py
```diff
@@ -634,6 +651,22 @@ def getRef(
...
+                linkType=linkType,
+                spec=spec,
+                status=status,
+                statusHint=statusHint,
+                linkFor=linkFor,
+                linkForHint=linkForHint,
+                explicitFor=explicitFor,
+                export=export,
+                ignoreObsoletes=True,
+                latestOnly=False)
```
nschonni, line 668:
```suggestion
                latestOnly=False,
            )
```

nschonni (commented):
Copying over the suggestions from the formatter

bikeshed/refs/ReferenceManager.py
fantasai: Thanks, fixed. :)

bikeshed/refs/ReferenceManager.py
```diff
@@ -519,6 +523,21 @@ def getRef(
             ignoreObsoletes=True,
         )
 
+        if failure and self.isDelta:
```
tabatkins, line 526: This block doesn't do what you want. `latestOnly` just finds, for each dfn from each shortname, the highest-level version and then suppresses lower-level version; if this is a delta spec that doesn't contain a term's dfn, it won't have a dfn in the results anyway so this won't have any effect.

(Also, it would only trigger if the first try completely failed; a confusable reference from another spec would prevent this from activating.)

tabatkins: What you instead want to alter is the ReferenceManager.removeSameSpecRefs() function; that's what unexports all dfns from specs with the same shortname.

(But make sure you still remove refs from the same or later level, so your local copy doesn't get confused with its version in the db, and doesn't link forward to a future version of itself either.)

tabatkins (approved):

PR #1600, opened 2020-01-24
Replace trailing with leading in documentation

### LINKED ISSUE #1488: Documentation about IDL keyword escaping says trailing, not leading
docs/index.bs:

> For example, IDL terms that match an IDL keyword must be defined in IDL with a trailing underscore, to avoid grammatical ambiguities, but actually define the term without the underscore. You can link to the term with either syntax.

But [Web IDL](https://heycam.github.io/webidl/#idl-names) sayeth:

> A leading "_" is used to escape an identifier from looking like a reserved word so that, for example, an interface named "interface" can be defined. The leading "_" is dropped to unescape the identifier.

Note that BS docs say "trailing" but Web IDL says "leading"

piyushpalawat99:
Currently, the documentation about IDL keyword says trailing which should be replaced with leading.
Closes #1488

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2098,7 +2098,7 @@ or by manually specifying the linking text on the <{a}>.
 
 In addition to English variations,
 these "conjugations" handle some IDL variations as well.
-For example, IDL terms that match an IDL keyword must be defined in IDL with a trailing underscore,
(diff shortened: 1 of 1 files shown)


tabatkins:
Thank you for noticing that!

PR #1602, opened 2020-02-13
Warn CGs who use WG statuses like ED.

jyasskin:
Partially fixes https://github.com/WICG/admin/issues/90 in Bikeshed. The other part is to notice CG repositories that are missing a `Group`.

@marcoscaceres


tabatkins:
Yup, looks good, thanks!

(I'm thinking these sets should actually move to a more visible config file in JSON at this point; intermixing code and data like this is getting unwieldy...)

------

So about detecting CGs that don't have a Group. Maybe it's enough to just test (1) no Group, and (2) repo name ends in "cg"? That catches all the ones you marked here, and none of the other groups.

I'm thinking I can log it as a warning, and allow you to silence it by explicitly specifying a `Group: none` (which currently doesn't work) if you aren't a CG but have a matching repo name for some reason.

We should also maybe put together a "w3ccg" Group with reasonable default boilerplate? Is that possible? I need to look at the different CG boilerplates and see how much difference there really is.

jyasskin:
All 4 of your paragraphs sound great. I suspect a w3ccg Group would be super helpful for new groups even if most of them just use it as a base for customization.

PR #1604, opened 2020-02-24
Include the W3C stylesheet in CG documents.

jyasskin:
I noticed that https://webbluetoothcg.github.io/web-bluetooth/ was missing the background we were trying to get from #1602, and it turns out that the WICG was the only group adding the right stylesheet. This PR fixes that problem, but it looks error-prone to hope all the CGs write their `header.include` correctly.


tabatkins:
Sigh, indeed. Thanks.

PR #1632, opened 2020-03-26
Set "Include Mdn Panels" for webapps boilerplate

sideshowbarker:


mkruisselbrink:
This seems to have broken specs that don't have mdn annotations yet? I.e. FileAPI is now failing on https://travis-ci.com/github/w3c/FileAPI/builds/155915673 with "FATAL ERROR: Couldn't find the MDN data for 'fileapi' nor 'fileapi'."

sideshowbarker:
> This seems to have broken specs that don't have mdn annotations yet? I.e. FileAPI is now failing on https://travis-ci.com/github/w3c/FileAPI/builds/155915673 with "FATAL ERROR: Couldn't find the MDN data for 'fileapi' nor 'fileapi'."

With https://github.com/tabatkins/bikeshed/commit/10e72f6737dc1fe4c290989e3c14e73014bf1894 there should no longer be build failures. See the updated docs at https://tabatkins.github.io/bikeshed/#metadata-include-mdn-panels:

> If your spec is not _yet_ in MDN, but you’d like the panels to be added when it shows up, you can pass one of the "maybe" values of a soft boolish to allow it to silently upgrade when possible; otherwise, if directly turned on, it will emit a fatal error if it can’t find your spec.

Apologies for the breakage caused by me not having thought to code it that way from the get-go.

PR #1635, opened 2020-03-27
Add WHATWG Commit Snapshot computed metadata

annevk:
@tabatkins api.csswg.org picks this up immediately, right? We need to change our invocation of that and I wonder how we need to balance things.

cc @domenic

### DIFF
--- a/bikeshed/boilerplate/whatwg/computed-metadata-LS-COMMIT.include
+++ b/bikeshed/boilerplate/whatwg/computed-metadata-LS-COMMIT.include
@@ -7,7 +7,7 @@
   ],
   "!Commits": [
     "<a href=https://github.com/whatwg/[SHORTNAME]/commits>GitHub whatwg/[SHORTNAME]/commits</a>",
-    "<a href=/ id=commit-snapshot-link>Go to the living standard</a>",
+    "<a href=https://[SHORTNAME].spec.whatwg.org/ id=commit-snapshot-link>Go to the living standard</a>",
     "<a href=https://twitter.com/[TWITTER]>@[TWITTER]</a>"
   ],
   "!Tests": "<a href=https://github.com/web-platform-tests/wpt/tree/master/[SHORTNAME]>web-platform-tests [SHORTNAME]/</a> (<a href=https://github.com/web-platform-tests/wpt/labels/[SHORTNAME]>ongoing work</a>)",
--- a/bikeshed/boilerplate/whatwg/computed-metadata-LS-COMMIT.include
+++ b/bikeshed/boilerplate/whatwg/computed-metadata-LS-COMMIT.include
@@ -11,5 +11,6 @@
     "<a href=https://twitter.com/[TWITTER]>@[TWITTER]</a>"
   ],
   "!Tests": "<a href=https://github.com/web-platform-tests/wpt/tree/master/[SHORTNAME]>web-platform-tests [SHORTNAME]/</a> (<a href=https://github.com/web-platform-tests/wpt/labels/[SHORTNAME]>ongoing work</a>)",
-  "Warning": "obsolete"
+  "Warning": "obsolete",
+  "Text Macro": "REPO-URL https://github.com/whatwg/[SHORTNAME]"
 }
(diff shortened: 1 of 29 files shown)


bikeshed/boilerplate/whatwg/computed-metadata-LS-COMMIT.include
```diff
@@ -0,0 +1,5 @@
+{
+  "Title": "[SPECTITLE] Standard Commit Snapshot [COMMIT-SHA]",
+  "Logo": "https://resources.whatwg.org/logo-[SHORTNAME]-snapshot.svg",
+  "!Commits": "<a href=/ id=commit-snapshot-link>Go to the living standard</a>"
```
domenic: Is it intentional to remove the other two links? I think it'd be reasonable to keep including them, as we do now.

bikeshed/boilerplate/whatwg/computed-metadata-LS-COMMIT.include
```diff
@@ -0,0 +1,5 @@
+{
+  "Title": "[SPECTITLE] Standard Commit Snapshot [COMMIT-SHA]",
```
domenic: This is a change (removing parentheses), but I guess it's a bit of an improvement.

bikeshed/boilerplate/whatwg/computed-metadata-LS-COMMIT.include
annevk: Yeah, it was intentional to include less information that might go stale in snapshots.

bikeshed/boilerplate/whatwg/computed-metadata-LS-COMMIT.include
annevk: This was also intended, to align Review Drafts and Commit Snapshots on a single style.

annevk:
Duplicating the whole header to remove the need for `md-warning="Commit $SHA $COMMIT_URL_BASE$SHA replaced by $LS_URL"` seems like the wrong direction to me.

bikeshed/boilerplate/whatwg/computed-metadata-LS-COMMIT.include
domenic: I mean, the information is a link to GitHub and Twitter... I'm not sure why those would go stale.

domenic:
Hmm. I like the way it keeps things centralized, but I see what you mean. There's probably also a way of avoiding the duplication, but I can try to make that work on top of this.

bikeshed/boilerplate/whatwg/computed-metadata-LS-COMMIT.include
annevk: I guess I don't really see a use case for them, that's why I also didn't include Tests/Participate.

domenic: Oh, those get deleted if you don't explicitly include them? I thought the defaults would apply.

I'm not super-excited about making the commit snapshots less usable in this way... they're meant to be nice stable references, and letting people navigate from them to the list of latest commits seems quite important, if they want to know e.g. what has changed since the snapshot they're looking at.

annevk: Oh, I haven't actually tested how that works. It might well work that way.

domenic:
OK, I:

- Added back the links that I prefer to keep, since in IRC @annevk said he'd be OK with that
- Inlined the logo instead of keeping it in a separate file; this removes a wrapper `<p>`. This has nothing to do with the rest of the PR and is just something that has been bugging me slightly. I'm happy to remove it from this PR.
- Managed to get automatic warnings generated into the right place using the `COMMIT-SHA`. https://github.com/whatwg/whatwg.org/pull/310 is updated to take this into account. It stops using the "commit" warning type in favor of the "obsolete" warning type, because the commit warning type wants specific inputs, whereas with the "obsolete" warning type we can just use the COMMIT-SHA input alone. This feels kind of like a hack but also kind of elegant.

We should probably not merge this until all three of us are online and can sync the whatwg.org PR and discuss whether it's the best strategy. But, I'm pretty happy with it.

annevk:
@domenic I checked with Tab and I can land this once we know what we want `.pr-preview.json` to look like. I also rebased the tests so CI will continue to pass.

domenic:
OK, this is ready. It adds a new status, `LS-PR`, specifically for PR previews. This allows us to get a dedicated warning and h1 that talk about the PR, for example.

annevk:
```
Update WHATWG metadata

To better account for Commit Snapshots and PR Preview.

See also https://github.com/whatwg/whatwg.org/pull/310 and https://github.com/whatwg/spec-factory/pull/5.
```

PR #1644, opened 2020-04-08
Use requests.get to replace urllib.request.urlopen

sideshowbarker:
The initial patch in this branch does a naïve replacement of all `urllib.request.urlopen` calls in the source with `requests.get`. It’s naïve in that it avoids changing the existing code any more than is minimally necessary to produce exactly the same behavior as the existing `urllib.request.urlopen` behavior — rather than making use of any additional benefits that using `requests` would bring, including automatic handling of gzipped responses.

This patch does at least prevent the problem of users in macOS XCode-installed Python3 environments running into SSL failures due to the current urllib code — because request installs and uses its own set of certs.

So if for right now we only want to first just replace the `urllib.request.urlopen` calls and solve the SSL problem without needing to alter the code any further, we could go ahead and merge this.

But if instead we also want to take time right now to change the code to make use of automatic handling of gzipped responses and other the additional benefits that switching to `requests` can bring, then we can do that instead.

Relevant details:

* `urllib.request.urlopen` by default sends no `Accept-Encoding` header in requests; thus, the responses the existing `urllib.request.urlopen` requests get back are never gzipped
* `requests.get` by default sends an `Accept-Encoding` header in requests, with `gzip` in the value, so the responses `requests.get` gets back may be gzipped
* `requests.get` will automatically gzip-decode responses if not used with `.raw`
* `requests.get(url, stream=True, headers=headers).raw` with `{"Accept-Encoding": "identity"}` is the functional-equivalent drop-in replacement for `urllib.request.urlopen`
* switching the code to use `requests.get` in way to get gzipped responses and automatically decode them will require more than just drop-in replacing the existing `urllib.request.urlopen` calls; the code surrounding/following the calls will also need to be touched (though the additional changes needed would not be all that big)

### DIFF
--- a/bikeshed/unsortedJunk.py
+++ b/bikeshed/unsortedJunk.py
@@ -7,7 +7,7 @@ import json
 import logging
 import os
 import re
-import urllib.request, urllib.parse, urllib.error
+import urllib.parse, urllib.error
 from collections import defaultdict, namedtuple
 from functools import partial as curry
 
--- a/bikeshed/update/manifest.py
+++ b/bikeshed/update/manifest.py
@@ -4,8 +4,7 @@
 import hashlib
 import io
 import os
-import urllib.request, urllib.error, urllib.parse
-from contextlib import closing
+import requests
 from datetime import datetime
 
 from ..messages import *
@@ -95,8 +94,7 @@ def updateByManifest(path, dryRun=False):
         warn("Couldn't find local manifest file.\n{0}", e)
         return False
     try:
-        with closing(urllib.request.urlopen(ghPrefix + "manifest.txt")) as fh:
-            remoteManifest = [str(line, encoding="utf-8") for line in fh]
+        remoteManifest = requests.get(ghPrefix + "manifest.txt").text.splitlines()
     except Exception as e:
         warn("Couldn't download remote manifest file.\n{0}", e)
         return False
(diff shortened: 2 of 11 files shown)


tabatkins:
Especially for the update code, we should go ahead and do gzip if at all possible; those files are *very* compressible.

Looking into it, it seems like Requests automatically decompresses for you, so if we just advertise that we accept gzip responses and *don't* access the raw response body, we'll get the same result but better?

sideshowbarker:
> Looking into it, it seems like Requests automatically decompresses for you, so if we just advertise that we accept gzip responses

Requests already does that automatically too — it always sends the following header:
```
Accept-Encoding: identity, deflate, compress, gzip
```
> and _don't_ access the raw response body, we'll get the same result but better?

Right. So it’s “just” a matter of changing the existing code to not access the raw response, since the current code is written to handle what `urllib.request.urlopen` returns — which is the same as the raw response body that `requests.get(…).raw` would return.

tabatkins:
If I'm reading it right, tho, we can just swap to grabbing `.text` (or `.json()` when we're just gonna immediately JSON-parse it), right?  I've definitely got no attachment to the precise mechanics of the current code. ^_^

sideshowbarker:
> If I'm reading it right, tho, we can just swap to grabbing `.text` (or `.json()` when we're just gonna immediately JSON-parse it), right?

Right. So yeah it shouldn’t take much time to make the changes — takes longer to test it to make sure nothing got broken/regressed inadvertently in any unsurprising ways.

Anyway I think I can make time today to try it. Then will push an update to this branch

sideshowbarker:
> we can just swap to grabbing `.text` (or `.json()` when we're just gonna immediately JSON-parse it), right?

OK, I force-pushed a change rewrites the change on this branch to do it all that way. 

I also changed my `json_home_client` locally, since that code had the same issues. I raised https://github.com/plinss/json_home_client/pull/3 with a patch to change that similarly.

I tested in my environment and confirmed that with those changes, I no longer get the *“CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate”* failure when using the Python 3.7 that ships with XCode.

PR #1648, opened 2020-04-16
Restore newlines when saving the remote manifest.

Ms2ger:
This was broken in #1644, as the elements of remoteManifest no longer have
traiing newlines.


sideshowbarker:
Thanks for catching this. I thought something pretty much exactly like this could likely happen with the #1644 change, but didn’t know how to test it. Guess I should have at least checked the output manually…

PR #1654, opened 2020-04-23
Add document expiration feature. Fixes #1652.

### LINKED ISSUE #1652: Document expiration
Some standards bodies have a notion of documents which expire at some point. Internet-Drafts at the IETF work this way, for instance, and @WICG may add one (see WICG/admin#102).

It'd be nice if Bikeshed supported this or could automate this somehow. Here are some handwavy thoughts:

* It should be possible to do this by hand for an individual document via metadata. e.g. add Expires and have it take the same format as Date.
* It should be possible for document statuses to get this automatically by default.
* The default bikeshed boilerplate should take this into account and throw up a warning (warning-expired.include), and this should be overridable / customizable per-group and per-spec like all other boilerplate is today

hober:
Here's a first cut at an expiration feature like we're talking about in #1652.

One annoyance is this adds a dependency on `isodate` in order to parse ISO 8601 durations.

### DIFF
--- a/bikeshed/metadata.py
+++ b/bikeshed/metadata.py
@@ -8,8 +8,9 @@ import os
 import re
 import subprocess
 from collections import defaultdict, OrderedDict
-from datetime import datetime
+from datetime import datetime, timedelta, date
 from functools import partial
+from isodate import parse_duration
 
 from . import attr
 from . import config
@@ -67,6 +68,7 @@ class MetadataManager:
         self.displayShortname = None
         self.editors = []
         self.editorTerm = {"singular": "Editor", "plural": "Editors"}
+        self.expires = None
         self.favicon = None
         self.forceCrossorigin = False
         self.group = None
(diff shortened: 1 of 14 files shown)


bikeshed/metadata.py
```diff
@@ -334,6 +345,35 @@ def parseDate(key, val, lineNum):
         return None
 
 
+def parseDateOrDuration(key, val, lineNum):
```
tabatkins, line 348: Also accept "never", returning None, so you can override a default expiry. Maybe allow a boolish false too?

tabatkins:
Add some tests and the one fix, and this is R+ from me

tabatkins:
Well, I guess we'll want to do the "starts closed, uses js to open after expiration" behavior for the not-yet-expired case

bikeshed/metadata.py
hober: Added never and boolish false in a33f871.

hober:
> Add some tests and the one fix, and this is R+ from me

Added tests and accept never & boolish false in a33f871.

Haven't done the JS bit yet.

tabatkins:
Okay, I added the JS and did some other minor fixup, including making sure the tests weren't dependent on the date they were generated, since that means they'll fail tomorrow. ^_^ Also added docs for the feature.

PR #1657, opened 2020-05-02
add code to verify requirements are installed

plinss:
TIL pkg_resources has code to parse and process requirements files, because, of course it does.

This code isn't part of a pip install, but will help out anyone running from source.

### DIFF
--- a/bikeshed.py
+++ b/bikeshed.py
@@ -1,4 +1,29 @@
 #!/usr/bin/env python
+
+def verify_requirements():
+    import os
+    import pkg_resources
+    requirements_file_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'requirements.txt')
+    if (os.path.exists(requirements_file_path)):
+        requirements_met = True
+        with open(requirements_file_path, 'r') as requirements_file:
+            requirements = [line for line in requirements_file.read().split('\n') if (not line.strip().startswith('-'))]
+            for requirement in pkg_resources.parse_requirements(requirements):
+                try:
+                    distribution = pkg_resources.get_distribution(requirement.project_name)
+                    if (distribution not in requirement):
+                        print('Package', requirement.project_name, 'version', distribution.version, 'is not supported')
+                        requirements_met = False
+                except Exception:
+                    print('Package', requirement.project_name, 'is not installed')
+                    requirements_met = False
+        if (not requirements_met):
+            print('Run "pip3 install -r {path}" to complete installation'.format(path=requirements_file_path))
+            exit()
+
+
+verify_requirements()
+
 import bikeshed
 
 if __name__ == "__main__":


hober:
This has caught me a couple of times. Landing this would save my brain a few steps next time.

tabatkins:
Yup, it caught me a few days ago too. One sec while I do a final review pass.

tabatkins:
And just cut `1.4.0` with the update. Had to fiddle with some stuff; `bikeshed.py` isn't run unless you do it manually; if you install it it just calls directly into `cli:main()` by itself.

plinss:
IIRC if you do a straight install via pip, the requirements.txt file isn’t included in the package. You’re probably going to have to include it in the package_data.

I originally but the check in Bikeshed.py partly due to this, and I figured pip would be maintaining the dependencies, tho they can be broken...

tabatkins:
`bikeshed.py` is never invoked if you use pip to install (either from pypi or locally). `setup.py` dictates the entry point for the `bikeshed` command, and it's straight to `main()`. The only purpose `bikeshed.py` has is letting you run `python bikeshed/bikeshed.py` to invoke it directly.

But good catch on `requirements.txt` not being packaged in, I've fixed that.

plinss:
> bikeshed.py is never invoked if you use pip to install (either from pypi or locally)

Right, which is why I originally put it in bikeshed.py and wrote:
> This code isn't part of a pip install, but will help out anyone running from source.

Totally cool to have it in __init__.py tho.

Did some more digging and found that the requirements are available via pkg_resources when you are installed via pip, so this could be done without the requirements.txt file when installed via pip. You won't have the requirements.txt to run pip with, but a plain `pip install bikeshed` should still fix broken dependencies. So having the requirements.txt in the package is kinda redundant.

I'll take a quick pass at writing a requirements.txt free version for when installed (what you have is fine, but I kinda want this for my own projects...)

plinss:
So, actually, it looks like when you're using the pip installed entry point it totally verifies the requirements for you at launch time, so no point in having this code anywhere other than bikeshed.py.

Since both you and @hober say you've been caught by, this I'm wondering if you were running via the pip entry point or bikeshed.py? (trying to understand this for my own sake)

tabatkins:
Like I said, if you do a local pip install (`pip install -e .`), it still uses the entry point defined in setup.py.  `bikeshed.py` isn't used at all by anything, unless you're manually executing that script yourself instead of using the pip-installed command.

pip does verify the requirements at launch time, but only the requirements it knows about; that is, the requirements that were present the last time you ran `pip install`. If you do a local install, and then later `git pull` to a version that has a new requirement, pip doesn't know about that, so you'll get the runtime error instead when you hit the missing `import`.

plinss:
Got it, thanks. The problem you and Tess ran into happens when the requirements are changed after install and you don't re-run `pip install -e .` to update them. Yeah, you'll need the version that always uses requirements.txt for that situation and can't rely on the pkg_resources info since that'll match what pip installed... but you probably don't need to include the requirements.txt in the package, because you'll only run into this issue when you have an editable install from source and the requirements.txt file will be there anyway. A normal pip install from pypi.org shouldn't run into the issue (and if it does, the entry point code will catch it).

So stick with what you have now.

tabatkins:
Ah that's true, only a local install needs it. So I'll be able to skip the check for pypi-sourced versions, which is good.

PR #1658, opened 2020-05-03
Gather all IDs when checking for required IDs

gsnedders:
This is a big speedup if there are lots of required IDs


tabatkins:
I hadn't noticed this in a trace yet, but probably just wasn't testing a spec with it; that's a classic N² right there I've had to clean up in the past. Thanks!

PR #1663, opened 2020-05-05
Clarify the status of the TAG, and provide TAG boilerplate.

hober:
Currently the TAG doesn't have any Bikeshed boilerplate and it isn't listed in the W3C megagroup, so TAG documents get Bikeshed's default boilerplate. Thus, Bikeshed produces TAG documents with the incorrect copyright, license, and SotD sections.

This PR fixes that, and cleans up the logic around W3C statuses more generally.

Note that I've updated this PR in 4a236e2d based on the conversation in w3c/AB-memberonly#35.


frivoal:
> The TAG can also publish REC-track documents.

Can it?  By my reading, the Process and the Patent Policy are clear to that only WGs can produce REC track documents (WGs and IGs can also produce Notes). The TAG is neither a WG or an IG. Certainly the tag can produce documents, but those aren't regulated by the Process, and cannot transition of official maturities granted by the Process.

Which of these maturities are you trying to use / have they been used by the TAG in the past?

hober:
@frivoal wrote:

> > The TAG can also publish REC-track documents.
> 
> Can it?

Yes. See the [Deliverables section of the TAG's charter](https://www.w3.org/2004/10/27-tag-charter.html#Deliverables1):

> The deliverables of the TAG are its Architectural Recommendations, review reports, and issue resolutions. The TAG may publish a variety of materials (e.g., short-term resolutions to issues that arise), but its Architectural Recommendations must be produced according to the formal Recommendation track process. As of the date of this charter, the TAG has produced one Recommendation track deliverable: [Architecture of the World Wide Web, First Edition](http://www.w3.org/TR/webarch/).

@frivoal wrote:

> Which of these maturities [have been] used by the TAG in the past?

At the minimum, WD, PR, and REC were used for [AWWW](https://www.w3.org/TR/webarch/).

hober:
I've updated this PR in 4a236e2d based on the conversation in w3c/AB-memberonly#35. Please take another look, @tabatkins @frivoal.

frivoal:
* Unlike RECs, I don't think there are any actual problem with it, but Notes are also for IGs and WGs only. In practice, I don't think there's a big difference between a Note and a Finding, is there? 
* The Status section of Findings talks about the Patent Policy. I don't think that's wrong, as the TAG is in the W3C, and all participants in W3C operate under the patent policy, but I'm also unsure it is useful, since the Patent Policy has nothing to say about Findings.

frivoal:
cc @wseltzer: What do you think about the second point in the previous comment? Should the status section of TAG findings include text referencing the Patent Policy? I don't think it does anything, so I'm inclined to drop it, but I'd rather have expert advice on that.

hober:
> * Unlike RECs, I don't think there are any actual problem with it, but Notes are also for IGs and WGs only.

The TAG has published at least WG Note recently:

https://www.w3.org/TR/security-privacy-questionnaire/

> * The Status section of Findings talks about the Patent Policy. I don't think that's wrong, as the TAG is in the W3C, and all participants in W3C operate under the patent policy, but I'm also unsure it is useful, since the Patent Policy has nothing to say about Findings.

Yeah, I'm not sure what would be best here either.

frivoal:
I can venture a guess, but I'd rather have it from you: what is the TAGs intention when it uses Findings for some of its work, and Notes for some other things?

hober:
> I can venture a guess, but I'd rather have it from you: what is the TAGs intention when it uses Findings for some of its work, and Notes for some other things?

I don't know. Maybe @plinss or @torgo have insights.

plinss:
In general, a Finding is a statement of the TAG, one step shy of a REC, but documenting an important architectural principle. These have historically been published in TAG space, but could in principle, live in /TR. They are treated as though they are published in /TR.

We use a Note when we see a value in having the document live in /TR space with persistent URL history for references (and is expected to be referenced by other /TR documents), but it's not generally documenting an architectural principal. e.g. the security and privacy questionnaire.

hober:
@frivoal, given https://github.com/w3c/AB-memberonly/issues/35#issuecomment-632910775, do you have any remaining concerns with this landing?

frivoal:
Not really, I think you can go ahead. We'll quite possibly have to revise this in a while, once we've figured out how to reconcile Process and practice, but in the meanwhile, the TAG should keep on TAGing, and that probably involves making bikeshed align with practice rather than with the Process.

PR #1671, opened 2020-05-20
Teach Bikeshed to accept a URL as its input source.

jyasskin:
There are a lot of debatable choices here, where I just picked one of several acceptable options. Let me know when a choice goes against your stylistic preferences. :)

### DIFF
--- a/bikeshed/InputSource.py
+++ b/bikeshed/InputSource.py
@@ -13,6 +13,7 @@ from typing import List, Optional
 import attr
 
 import requests
+import retrying
 
 from .Line import Line
 
@@ -109,11 +110,26 @@ class UrlInputSource(InputSource):
     def __str__(self) -> str:
         return self.sourceName
 
-    def read(self) -> InputContent:
-        response = requests.get(self.sourceName)
-        if response.status_code != 200:
-            # This matches the OSErrors expected by older uses of FileInputSource.
+    @retrying.retry(
+        stop_max_attempt_number=3,
+        wait_fixed=1000,
+        # Only catch exceptions from the requests library.
+        retry_on_exception=lambda e: isinstance(
+            e, requests.exceptions.RequestException
+        ),
+    )
+    def _fetch(self, *args, **kwargs):
+        response = requests.get(self.sourceName, timeout=10)
+        if response.status_code == 404:
+            # This matches the OSErrors expected by older uses of
+            # FileInputSource. It skips the retry, since the server has given us
+            # a concrete, expected answer.
             raise FileNotFoundError(errno.ENOENT, response.text, self.sourceName)
+        response.raise_for_status()
+        return response
+
+    def read(self) -> InputContent:
+        response = self._fetch(self.sourceName)
         date = None
         if "Date" in response.headers:
             # Use the response's Date header, although servers don't always set
--- a/bikeshed/InputSource.py
+++ b/bikeshed/InputSource.py
@@ -1,0 +1,155 @@
+from __future__ import annotations
+
+import email.utils
+import errno
+import io
+import os
+import sys
+import urllib.parse
+from abc import abstractmethod
+from dataclasses import dataclass
+from datetime import date, datetime
+from typing import List, Optional
+
+import requests
+
+from .Line import Line
+
+
+@dataclass
+class InputContent:
+    rawLines: List[str]
+    date: Optional[date]
+
+    @property
+    def lines(self) -> List[Line]:
+        return [Line(i, line) for i, line in enumerate(self.rawLines, 1)]
+
+    @property
+    def content(self) -> str:
+        return "".join(self.rawLines)
+
+
+class InputSource:
+    """Represents a thing that can produce specification input text.
+
+    Input can be read from stdin ("-"), an HTTPS URL, or a file. Other
+    InputSources can be found relative to URLs and files, and there's a context
+    manager for temporarily switching to the directory of a file InputSource.
+    """
+
+    def __new__(cls, sourceName: str):
+        """Dispatches to the right subclass."""
+        if cls != InputSource:
+            # Only take control of calls to InputSource(...) itself.
+            return super().__new__(cls)
+
+        if sourceName == "-":
+            return StdinInputSource(sourceName)
+        elif sourceName.startswith("https:"):
+            return UrlInputSource(sourceName)
+        else:
+            return FileInputSource(sourceName)
+
+    @abstractmethod
+    def __str__(self) -> str:
+        pass
+
+    @abstractmethod
+    def read(self) -> InputContent:
+        """Fully reads the source.
+        """
+        pass
+
+    def hasDirectory(self) -> bool:
+        """Only some InputSources have a directory."""
+        return False
+
+    def directory(self) -> str:
+        """Suitable for passing to subprocess(cwd=)."""
+        raise TypeError("{} instances don't have directories.".format(type(self)))
+
+    def relative(self, relativePath) -> Optional[InputSource]:
+        """Resolves relativePath relative to this InputSource.
+
+        For example, InputSource("/foo/bar/baz.txt").relative("quux/fuzzy.txt")
+        will be InputSource("/foo/bar/quux/fuzzy.txt").
+
+        If this source type can't find others relative to itself, returns None.
+        """
+        return None
+
+    def mtime(self) -> Optional[float]:
+        """Returns the last modification time of this source, if that's known.
+        """
+        return None
+
+
+class StdinInputSource(InputSource):
+    def __init__(self, sourceName: str):
+        assert sourceName == "-"
+        self.type = "stdin"
+        self.sourceName = sourceName
+        self.content = None
+
+    def __str__(self) -> str:
+        return "-"
+
+    def read(self) -> InputContent:
+        return InputContent(sys.stdin.readlines(), None)
+
+
+class UrlInputSource(InputSource):
+    def __init__(self, sourceName: str):
+        assert sourceName.startswith("https:")
+        self.sourceName = sourceName
+        self.type = "url"
+
+    def __str__(self) -> str:
+        return self.sourceName
+
+    def read(self) -> InputContent:
+        response = requests.get(self.sourceName)
+        if response.status_code != 200:
+            # This matches the OSErrors expected by older uses of FileInputSource.
+            raise FileNotFoundError(errno.ENOENT, response.text, self.sourceName)
+        date = None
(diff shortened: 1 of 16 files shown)


bikeshed/InputSource.py
```diff
@@ -0,0 +1,155 @@
+from __future__ import annotations
```
jyasskin, line 1: Improves the syntax for type annotations of methods like `relative()` that return the type of the class they appear inside of. I can drop this and name those types with a string.

https://www.python.org/dev/peps/pep-0563/

bikeshed/InputSource.py
```diff
@@ -0,0 +1,155 @@
...
+from dataclasses import dataclass
+from datetime import date, datetime
+from typing import List, Optional
+
+import requests
+
+from .Line import Line
+
+
+@dataclass
```
jyasskin, line 19: Very similar to attr, but:

1. attr grew support for mypy in version 18.2.0, but bikeshed embeds (?!?) 17.3.0.
2. [dataclass doesn't support slots](https://www.python.org/dev/peps/pep-0557/#support-for-automatically-setting-slots). :(
3. PyRight, the VS Code default python formatter/typechecker, [doesn't intend to support attr](https://github.com/microsoft/pyright/issues/146).

Probably the right thing to do is to import a newer attr from PyPi, which I'll send you another PR to do.

bikeshed/InputSource.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def __str__(self) -> str:
+        return self.sourceName
+
+    def read(self) -> InputContent:
+        with io.open(self.sourceName, "r", encoding="utf-8") as f:
+            return InputContent(
+                f.readlines(),
+                datetime.fromtimestamp(os.path.getmtime(self.sourceName)).date(),
+            )
```
jyasskin, line 141: I formatted this file with [black](https://black.readthedocs.io/en/stable/), but it might have different style opinions than you do... Which formatter do you use?

bikeshed/Spec.py
```diff
@@ -262,11 +260,12 @@ def serialize(self):
     def fixMissingOutputFilename(self, outputFilename):
         if outputFilename is None:
             # More sensible defaults!
-            if self.inputSource.endswith(".bs"):
-                outputFilename = self.inputSource[0:-3] + ".html"
-            elif self.inputSource.endswith(".src.html"):
-                outputFilename = self.inputSource[0:-9] + ".html"
-            elif self.inputSource == "-":
+            if isinstance(self.inputSource, FileInputSource):
```
jyasskin, line 263: We could move the output filename guessing into the InputSource class, and we could use something other than "-" for URLs, like use the last segment of the URL's path.

jyasskin:
Oh, my other open question is about the number of implicit requests this will cause to whatever server hosts the spec source. Adding some diagnostic output to `bikeshed.py spec https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/index.bs index.html` shows:

```
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/index.bs
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/defaults-ED.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/defaults.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/computed-metadata-ED.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/computed-metadata.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/bs-extensions-ED.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/bs-extensions.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/anchors.bsdata
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/link-defaults.infotree
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/header-ED.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/header.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/footer-ED.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/footer.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/xsite-tracking-model.bsinc
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/capabilities.bsinc
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/goals.bsinc
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/essential.bsinc
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/status-ED.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/status.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/logo-ED.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/logo.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/copyright-ED.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/copyright.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/abstract-ED.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/abstract.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/warning-not-ready-ED.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/warning-not-ready.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/stylesheet-ED.include
Fetching https://raw.githubusercontent.com/w3cping/privacy-threat-model/main/stylesheet.include
```

of which 80% are boilerplate includes I'm not actually using. Is it worth doing something to reduce that number? Or should we do something only when it becomes a problem?

These serialized fetches also take about 6.5s of extra wall time for this particular spec.  I feel like that's acceptable, and fixing it would require some more rearchitecting to make all the I/O asynchronous. The good news is that Python 3.5 added [general async/await support](https://www.python.org/dev/peps/pep-0492/), so there's at least a framework for doing that.

bikeshed/InputSource.py
tabatkins: #1672 is merged, so feel free to swap this back over to attr

bikeshed/InputSource.py
tabatkins: I don't! Probably should at some point, but for now it's hand-formatted. I'm fine with these decisions.

bikeshed/InputSource.py
```diff
@@ -0,0 +1,155 @@
...
+import email.utils
+import errno
+import io
+import os
+import sys
+import urllib.parse
+from abc import abstractmethod
+from dataclasses import dataclass
+from datetime import date, datetime
+from typing import List, Optional
```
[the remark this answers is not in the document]
tabatkins: Oh def not, I'm happy to get started on annotations.

bikeshed/InputSource.py
tabatkins: Looking over the PEP, I'm fine with this.

tabatkins:
I feel like boilerplate includes, since they're completely ambient (nothing in the source indicates you're using them) and quite numerous, should probably be omitted from this feature unless you explicitly opt in. Same with the datafile includes.  (Both of these are also rarely *used* as far as I can tell.)

So probably this should sprout a metadata that lists the names of boilerplate files to fetch, or lets you opt into all of them easily; we can make that pretty easy by setting the metadata value to a BoolSet (so you can set the default to either True or False, depending).

Bonus points to grow a new `bikeshed source` option that monitors for local boilerplate/datafile usage, and inserts the appropriate metadata automatically.

bikeshed/Spec.py
tabatkins: What would that mean? We do a POST to that URL?

bikeshed/InputSource.py
jyasskin: Done.

bikeshed/InputSource.py
```diff
@@ -0,0 +1,155 @@
...
+        self.sourceName = sourceName
+        self.type = "url"
+
+    def __str__(self) -> str:
+        return self.sourceName
+
+    def read(self) -> InputContent:
+        response = requests.get(self.sourceName)
+        if response.status_code != 200:
+            # This matches the OSErrors expected by older uses of FileInputSource.
```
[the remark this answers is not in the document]
jyasskin: Thanks for the pointer. I avoided retrying 404s, and there may be some more 4xx codes we should just pass up, but I'm not sure what they are yet.

bikeshed/Spec.py
jyasskin: No, I was thinking to save it to the current directory, using a name based on the last segment. I'm happy with the '-' default for now.

bikeshed/Spec.py
```diff
@@ -262,11 +260,12 @@ def serialize(self):
...
+                if self.inputSource.sourceName.endswith(".bs"):
+                    outputFilename = self.inputSource.sourceName[0:-3] + ".html"
+                elif self.inputSource.endswith(".src.html"):
+                    outputFilename = self.inputSource.sourceName[0:-9] + ".html"
```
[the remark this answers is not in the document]
jyasskin: Oops, fixed.

bikeshed/config/retrieve.py
```diff
@@ -76,26 +77,22 @@ def boilerplatePath(*segs):
...
-    filenames.append(boilerplatePath(genericFile))
+        sources.append(InputSource(boilerplatePath(group, statusFile)))
+        sources.append(InputSource(boilerplatePath(group, genericFile)))
+    sources.append(InputSource(boilerplatePath(statusFile)))
+    sources.append(InputSource(boilerplatePath(genericFile)))
 
-    for filename in filenames:
-        if os.path.isfile(filename):
+    for source in sources:
+        if source is not None:
```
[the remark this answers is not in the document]
jyasskin: `source` is `None` if `.relative()` isn't supported, so for the `StdinFileSource`.

jyasskin:
I've added the metadata, with a warning when it can locally determine you've forgotten an entry, but I didn't go for the `bikeshed source` bonus points. PTAL. :)

tabatkins:
Nice, that works for me.

PR #1672, opened 2020-05-20
Update attrs from 17.3.0.dev0 to 19.3.0.

jyasskin:
All of the changes in the test suite are clearly wrong, but I have no idea why they're happening. Downgrading my local install of attrs to 17.3.0 doesn't get rid of them. I can dig more tomorrow, but I wonder if you can just see what's happening here?


jyasskin:
Oh hey, the errors here match the TravisCI failures at head, so they're not my fault. :)

tabatkins:
I actually just plain forgot that that I'd vendored attr; thanks for swapping it out. ^_^

PR #1682, opened 2020-05-28
Add boilerplate for Browser Testing and Tools Working Group

foolip:
WD matches https://www.w3.org/TR/webdriver/
ED matches wording and order of many other status-ED.include


foolip:
This is to enable https://github.com/w3c/webdriver/issues/1462.

However, I have my doubts that LD is a status we should use here. It results in https://www.w3.org/StyleSheets/TR/2016/W3C-LD being used as stylesheet, and that's 404.

@AutomatedTester @sideshowbarker what should be the statuses of tip-of-tree https://w3c.github.io/webdriver/ and https://w3c.github.io/webdriver-bidi/? I'd like to include the template text for those statuses at least.

foolip:
My best guess is Editor's Draft and Unofficial Draft, respectively.

sideshowbarker:
> @AutomatedTester @sideshowbarker what should be the statuses of tip-of-tree https://w3c.github.io/webdriver/

That could be considered WD since the same revision of the document is also auto-published to https://www.w3.org/TR/webdriver/

> and https://w3c.github.io/webdriver-bidi/?

Personally, I would label it Editor’s Draft. I don’t know what the difference between Editor’s Draft and Unofficial Draft is supposed to be, but as far as I can see, there is no policy at the W3C for labeling anything as an Unofficial Draft — and we have many precedents of the label Editor’s Draft being used for documents which haven’t yet been formally adopted as deliverables by any working groups, nor published yet as Working Drafts.

AutomatedTester:
For context, webdriver is that we want a living doc and it relies on the evergreen work that @plehegar and @fantasai are/were working on.

tabatkins:
Yeah, UD is (unofficially, appropriate enough) often used as an "even less official" editor's draft, for things that shouldn't be read as having *any* blessing from the WG.

> However, I have my doubts that LD is a status we should use here. It results in w3.org/StyleSheets/TR/2016/W3C-LD being used as stylesheet, and that's 404.

Whatever the stylesheet is that *should* be used for LDs, we can fix that in bikeshed. What URL shoudl it use?

foolip:
I've updated to have WD and ED templates instead now. Note that this depends on https://github.com/w3c/webdriver/pull/1511 for the "WD matches https://www.w3.org/TR/webdriver/" claim in the description to be 100% true, as the placement of `[STATUSTEXT]` is different, in order to match most other templates and ED.

foolip:
@AutomatedTester, @sideshowbarker and I discussed a bit on IRC and concluded that the ED+WD boilerplate is OK, no need for LD. I'll squash the PR to clean up the history.

tabatkins:
No need to squash the PR, btw; I squash at the merge anyway.

PR #1689, opened 2020-06-01
Fix #1687: Watch includes

### LINKED ISSUE #1687: `bikeshed watch` doesn't rebuild on changes to included files
As should be expected since https://github.com/tabatkins/bikeshed/blob/master/bikeshed/Spec.py#L344 only looks at `self.inputSource` and not anything it includes.

jyasskin:
This attempt is the simplest thing that could possibly work. It doesn't cover watching boilerplate files and abandons trying to print which file changed.

### DIFF
--- a/bikeshed/InputSource.py
+++ b/bikeshed/InputSource.py
@@ -57,6 +57,9 @@ class InputSource:
     def __str__(self) -> str:
         pass
 
+    def __repr__(self) -> str:
+        return "{}({!r})".format(self.__class__.__name__, str(self))
+
     def __hash__(self):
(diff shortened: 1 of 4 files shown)


tabatkins:
Looks good overall. Only comment is that I don't like `assembleDocument()` using its return value to communicate this info; I'd prefer it be stored on a new `Spec` attribute instead.

jyasskin:
Done, which also made it straightforward to watch boilerplate dependencies.

tabatkins:
Ah, that's very nice now. Thanks!

PR #1691, opened 2020-06-03
Test Bikeshed with both Python 3.7 and 3.8.

jyasskin:
This doesn't test 3.9-dev since it's not worth the build spam for any failures
while they're beta-testing.

Since #1690 tells users to use 3.8, the automated build should probably flag if anything stops working there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 python:
  - "3.7"
+ - "3.8"
 before_install:
  - pip install flake8
 install:


tabatkins:
I did not realize you could do that!

jyasskin:
Yep, the example in https://docs.travis-ci.com/user/languages/python/#specifying-python-versions literally works.

https://docs.travis-ci.com/user/build-matrix/ has more details including the ability to [let `3.9-dev` run but not fail the whole build](https://docs.travis-ci.com/user/build-matrix/#rows-that-are-allowed-to-fail).

foolip:
Good call, I didn’t think to check what the CI setup of Bikeshed itself was.

Issue #1695, opened 2020-06-05
Have bikeshed help with publication announcements


frivoal:
When pushing out a new publication to TR, it's good practice to send announcements by email, to blog about them, to tweet about them, etc. But it's annoying, so people tend to skip.

However, if you look at the templates proposed in https://wiki.csswg.org/spec/publish, bikeshed already knows (most of the) information that should go in there. For the mail version, it's something like

> The [[Group]] has published a new [[Status]] of [[Title]]:
> 
>     [[TR]]
> 
> [[Abstract]]
> 
> [[Status Text]]
> 
> See also the change section for details:
>     [[This Version]]#changes
> 
> Please review the draft, and send any comments to this mailing list,
> <www-style@w3.org>, prefixed with [[Shortname]] (as I did on this
> message) or (preferably) file them in the GitHub repository at
>   https://github.com/w3c/csswg-drafts/issues

And for the blog version, it's a bit different thanks to being able to use markup, but not all that different.

Granted, this assumes you're keeping the `Status Text` and `Abstract` up to date, but you should do that anyway.

If something like `bikeshed echidna --with-announcements` could generate that text for me, that'd be cool. If it could (given credentials) post it for me, that would be even cooler.

fantasai:
The point of the announcement is not just to say "we did a thing, go look at it", but to actually communicate something interesting about the status. So I'm not sure automatically posting is a good idea even though it sounds nice. It's good for the author to actually think what's relevant commentary to post.

frivoal:
I agree that the author should do that, but communicating something interesting about the status of the spec is also what the status section of the spec is supposed to do, and most of the time it's neglected. I'm thinking that by tying this together, editors will get an incentive to write their `Status Text`, and get a mail/blog for free when they do that.

frivoal:
And besides, if you think the spec's status text section should be different from the mail's "something interesting about the status", the tool can take that into account, and draw it from some other source, and still save some effort composing the mail, escaping the things, replacing the bits in the template, etc.

fantasai:
@frivoal Agree the Status section should say something interesting, but it's so much boilerplate atm it's not very useful even when it does. :( Wrt saying something different... I usually have slightly different text for announcements depending on the audience, and some announcements are CCed to e.g. www-international and some aren't etc. So there's often some amount of tweaking in my experience.

frivoal:
[...]

PR #1698, opened 2020-06-05
Rough initial draft of bikeshed's announcement features

frivoal:
This is a rough sketch of what solving https://github.com/tabatkins/bikeshed/issues/1695 might look like.

This is clearly incomplete:
* the code badly needs cleaning up
* I haven't written the wordpress curl part
* `[ABSTRACT]` and `[STATUSTEXT]` generate a `<p line-number=-1>` / `</p>` wrapper that I don't want
* HTML markup should be stripped from the email body (and `<a href="https://foo">link</a>` should probably be turned into something like `link (see https://foo)`

Still, it's a start.

### DIFF
--- a/bikeshed/boilerplate/csswg/blog.include
+++ b/bikeshed/boilerplate/csswg/blog.include
@@ -1,0 +1,10 @@
+Title: New [STATUS] of [TITLE]
+Categories: Publications
+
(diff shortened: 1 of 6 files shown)


frivoal:
I thought this could be nice, but others don't seem interested, and I don't have time to maintain a fork or polish this up.

PR #1710, opened 2020-06-19
Enable MDN panels for all WHATWG specs

sideshowbarker:
Depends on https://github.com/whatwg/whatwg.org/pull/336 and https://github.com/whatwg/whatwg.org/pull/338


bikeshed/boilerplate/whatwg/footer.include
```diff
@@ -10,3 +10,4 @@ if ("serviceWorker" in navigator) {
   navigator.serviceWorker.register("/service-worker.js");
 }
 </script>
+<script src="https://resources.whatwg.org/standard-mdn-annos.js"></script>
```
domenic, line 13: I think the bottom of the page is probably not the right place for this. I'd suggest at the top, but with the `defer=""` attribute so that it executes at the bottom. That way it gets fetched ASAP.

sideshowbarker: > I think the bottom of the page is probably not the right place for this. I'd suggest at the top, but with the `defer=""` attribute so that it executes at the bottom. That way it gets fetched ASAP.

OK, thanks — moved it to the sop

sideshowbarker:
With whatwg/whatwg.org#336 and whatwg/whatwg.org#338 now both merged — and with https://resources.whatwg.org/standard-mdn-annos.js now deployed — I believe this PR is now ready to be merged.

bikeshed/boilerplate/whatwg/defaults.include
```diff
@@ -5,7 +5,8 @@
   "Force Crossorigin": "yes",
   "Status": "LS",
   "No Editor": "yes",
-  "Boilerplate": "omit style-md-lists, omit style-autolinks, omit style-selflinks, omit style-counters, omit style-syntax-highlighting, omit feedback-header, omit conformance, omit issues-index",
+  "Boilerplate": "omit style-md-lists, omit style-autolinks, omit style-selflinks, omit style-counters, omit style-syntax-highlighting, omit feedback-header, omit conformance, omit issues-index, style-mdn-anno no, script-mdn-anno no",
```
domenic, line 8: Should this be "omit" prefix instead of "no" suffix?

sideshowbarker: Thanks — yeah, changed to "omit" for consistency

PR #1726, opened 2020-06-30
WHATWG: separate IPR

annevk:
As per https://github.com/whatwg/sg/issues/124.

@tabatkins is this enough to go on? Basically, if "Latest RD" or some such metadata is available the conditional bit should end up saying:

```html
<p>This is the Living Standard. Those interested in the patent-review version should view the 
<a href="/review-drafts/[Review RD]/">Living Standard Review Draft</a>.</p>
```

### DIFF
--- a/bikeshed/boilerplate/whatwg/defaults-RD.include
+++ b/bikeshed/boilerplate/whatwg/defaults-RD.include
@@ -9,6 +9,6 @@
   "Informative Classes": "domintro",
   "Metadata Include": "Translations no",
   "Informative Classes": "domintro",
-  "Tracking Vector Image": "https://resources.whatwg.org/tracking-vector.svg"
+  "Tracking Vector Image": "https://resources.whatwg.org/tracking-vector.svg",
   "Slim Build Artifact": "yes"
 }
--- a/bikeshed/boilerplate/whatwg/footer-RD.include
+++ b/bikeshed/boilerplate/whatwg/footer-RD.include
@@ -1,3 +1,7 @@
+<h2 id="ipr" class="no-num">Intellectual property rights</h2>
+
+<div data-fill-with="ipr" include-if="boilerplate:ipr"></div>
+
 <p>Copyright © WHATWG (Apple, Google, Mozilla, Microsoft). This work is licensed under a
 <a rel="license" href="https://creativecommons.org/licenses/by/4.0/">Creative Commons Attribution
 4.0 International License</a>.
--- a/bikeshed/boilerplate/whatwg/footer.include
+++ b/bikeshed/boilerplate/whatwg/footer.include
@@ -1,6 +1,14 @@
+<h2 id="ipr" class="no-num">Intellectual property rights</h2>
+
+<div data-fill-with="ipr" include-if="boilerplate:ipr"></div>
+
 <p>Copyright © WHATWG (Apple, Google, Mozilla, Microsoft). This work is licensed under a
 <a rel="license" href="https://creativecommons.org/licenses/by/4.0/">Creative Commons Attribution
 4.0 International License</a>.
+
+<p include-if="text macro: LATESTRD">This is the Living Standard. Those interested in the
+patent-review version should view the
+<a href="/review-drafts/[LATESTRD]/">Living Standard Review Draft</a>.</p>
 
 </main>
 
--- a/tests/github/whatwg/compat/compatibility.html
+++ b/tests/github/whatwg/compat/compatibility.html
@@ -139,6 +139,7 @@
        </ol>
      </ol>
     <li><a href="#acknowledgements"><span class="secno"></span> <span class="content">Acknowledgements</span></a>
+    <li><a href="#ipr"><span class="secno"></span> <span class="content">Intellectual property rights</span></a>
     <li>
      <a href="#index"><span class="secno"></span> <span class="content">Index</span></a>
      <ol class="toc">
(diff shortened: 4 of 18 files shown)


annevk:
@domenic @foolip @tabatkins what do we want to do if a specification needs to acknowledge the rights of something that got used similar to https://html.spec.whatwg.org/multipage/acknowledgements.html#ipr? If we anticipate that happening, it would make more sense for specifications to include an IPR section themselves and let part of it be filled through Bikeshed's boilerplate system, right? Or do we have more options such as only resorting to that if it includes an IPR section of its own and otherwise letting Bikeshed handle it in its entirety?

foolip:
@annevk since we're not sure it will happen, I'd take the approach of the whole section being possible to suppress and done manually, rather than adding a slot for extra stuff. It should be similar to abstracts, then.

annevk:
@tabatkins thinking about it more (and chatting with Philip) what would be ideal here is being able to emulate https://html.spec.whatwg.org/multipage/acknowledgements.html#ipr. So:

1. We put the heading in the boilerplate.
2. Then there's a conditional multi-line markup insert for attribution.
3. Then there's the copyright line.
4. Then there's a conditional RD line insert as per above.

Would you be willing to build this or advice on an approach?

tabatkins:
So here's a quick design for conditional inclusion; let me know if it meets your needs (I *think* it does):

* any element can have an `include-if` and/or an `exclude-if` attribute, containing a comma-separated list of conditions
* if an element has `include-if`, it's included only if it matches at least one condition; if it has `exclude-if` it's included only if it matches none of the conditions
* conditions can be statuses, or `!Foo: bar` custom metadata, with exact-match semantics after trimming whitespace

This work?

tabatkins:
Probably also have a container whose *sole* purpose is to include/exclude its contents, and which itself is always automatically removed from the output. (But if its checks pass, its *contents* get lifted up in place of itself.)

annevk:
I'm not sure if it helps with 2, but I suppose in theory we could put multi-line markup in custom metadata and that would work. That is, if
```
<h2>...</h2>

<p>...</p>
```
is part of our boilerplate, how would we allow a specification to add additional text there between those two elements?

tabatkins:
While you *could* smuggle it all into `Text Macro` values, an easier way would be to put a `<div data-fill-with=whatwg-custom-stuff include-if=RD></div>` into the boilerplate, and then have RDs include a `<div boilerplate=whatwg-custom-stuff>...</div>` element with all the markup you want. It'll get reslotted into the spot in the boilerplate you designated for it.

tabatkins:
All right, new conditional feature, and I documented custom boilerplate sections a little better.  I think that enables you to solve this; please let me know if there's anything confusing or lacking.

tabatkins:
Hm, looks like from your first comment you want the ability to make conditionals based on a text macro being defined. Is that needed here?

annevk:
@tabatkins `<div boilerplate>` looks like it would do the trick, yes!

Now, for the conditional, this is for inclusion in the Living Standard, _if_ it has a review draft. Currently all our standards have a Review Draft, but if we spin up a new one, it might not and then it should not have this message.

What I was thinking is that existence of the boilerplate key could be used. So:
```
<p include-if="Review RD">This is the Living Standard. Those interested in the patent-review version 
should view the <a href="/review-drafts/[Review RD]/">Living Standard Review Draft</a>.</p>
```

tabatkins:
That example is for a `Text Macro` value, not a `boilerplate=""`, right?

Both sound very reasonable and have an obvious use-case. Okay, so I'm thinking a key like `include-if="boilerplate: foo"` or `include-if="text macro: Review RD"`

annevk:
@tabatkins for in `<pre class=metadata>`, e.g., `Review RD: 2005-12`. (I think that means yes.)

The `boilerplate=""` case I'm thinking of is all the text in https://html.spec.whatwg.org/multipage/acknowledgements.html#ipr below the heading and before the copyright line. Most standards don't need that, but if they do it would be nice if it could be formatted that way while still letting the heading and copyright notice be managed centrally here.

Hope that helps.

tabatkins:
Okay, I've added (and documented) both Text Macro and boilerplate inclusion conditions.

I am slightly confused by your examples still, but I assume it's because you're writing them off the cuff rather than taking from the actual spec. There's no `Review RD` metadata; a `!Review RD` metadata might exist, but that just adds the line to the spec-metadata block (along with editors, status, etc), it doesn't create a text macro; text macros must be all-caps and a single word, so `Text Macro: Review RD 2005-12` wouldn't work (but `Text Macro: REVIEWRD 2005-12` would).

Assuming the final thing if what you're referring to, then yes, this would work:

```
<p include-if="text macro: REVIEWRD">This is the Living Standard. Those interested in the patent-review version 
should view the <a href="/review-drafts/[REVIEWRD]/">Living Standard Review Draft</a>.</p>
```

and would only include this paragraph if that macro is defined.

----

For the IPR boilerplate case, it looks like you could get away with just using a `<div data-fill-with=ipr></div>` element in the boilerplate, and if specs include a `<div boilerplate=ipr>...</div>` it'll get moved there. If they don't have it, the `<div data-fill-with>` will stick around in the output, but it'll be empty and so probably okay. But if you amend it to `<div data-fill-with=ipr include-if="boilerplate:ipr"></div>`, then it'll remove itself from the output when there's nothing to fill it with, too.

annevk:
I see, my apologies, `Review RD` isn't something that exists yet and a `Text Macro` does indeed seem like it would fit the bill. I'll play with this tomorrow. Thanks so much!

PR #1727, opened 2020-07-07
Use <div> rather than <aside> for MDN panels

sideshowbarker:
This change makes Bikeshed emit `<div>` elements for MDN panels, rather than `<aside>` elements.

Otherwise, without this change, some markup that Bikeshed emits for MDN panels doesn’t conform to current HTML spec requirements. Specifically, the HTML spec currently doesn’t allow `<aside>` elements as descendants of `<dt>` elements — but we are intentionally inserting MDN panels as `<dt>` descendants (to get them aligned at point of use). So because of that, for now at least we need to use `<div>` for the panels, rather than `<aside>`.

Relates to https://github.com/whatwg/wattsi/issues/129

### DIFF
--- a/bikeshed/mdnspeclinks.py
+++ b/bikeshed/mdnspeclinks.py
@@ -95,7 +95,7 @@ def addMdnPanels(doc):
             '''  # noqa
 
 def createAnno(className, mdnButton, featureDivs):
-    return E.aside({"class": className}, mdnButton, featureDivs)
+    return E.div({"class": className}, mdnButton, featureDivs)
 
 def panelsFromData(doc, data):
     mdnBaseUrl = "https://developer.mozilla.org/en-US/docs/Web/"


tabatkins:
I think this is part of why I originally put them at the end of the document - `aside` is the correct element, semantically, but it couldn't go everywhere it needed to go.

Seems like maybe we should just revisit the restriction against aside-in-dt?

tabatkins:
Or I guess as Domenic suggested, we can move them to the following `dd` instead.

domenic:
Ping on getting this merged or solved in some way sooner rather than later. Since https://github.com/tabatkins/bikeshed/pull/1710 was merged, WHATWG specs are now failing to build, as we do not allow deploying without passing the HTML checker.

tabatkins:
I'll go ahead and merge this, but it's not the correct solution - the dfn panels *are* asides.

sideshowbarker:
> Since #1710 was merged, WHATWG specs are now failing to build, as we do not allow deploying without passing the HTML checker.

Sorry, I should have anticipated that.

> Seems like maybe we should just revisit the restriction against aside-in-dt?

I think we should, as I alluded to in https://github.com/whatwg/wattsi/issues/129#issuecomment-654661017. I think the spec restriction that disallows developers/authors from using `aside` within `dt` is a case where the spec is being overly prescriptive.

It’s not clear to me what real problem for end users and developers that no-aside-within-dt restriction is helping to prevent. I believe it’s instead unnecessarily disallowing some legitimate use cases — as I pointed out in https://github.com/whatwg/wattsi/issues/129#issuecomment-654661017:

> Given that aside began life as a basically a synonym for footnote, and that it’s easily imaginable that authors/developers may want, for example, to footnote dt content (which is essentially what we are doing with the MDN annos for dt content), then I don’t think we should be preventing/discouraging authors/developers from being able to do that.
>
> So it’s not clear to me what problem we are trying to help avoid authors/developers avoid by making aside within dt non-conforming. Unless it’s clear to others what the problem is, then I suggest we consider dropping that restriction.

As far as the alternative of moving the panels/annotations to the following `dd` instead, I think that should be the least-preferred solution at the point. I think we should only reconsider that after having a discussion about the spec issue.

But if after that discussion, it ends up being clear that the spec restriction really does in fact make the best sense for end users and for developers, then I agree we should move the panels/annotations to the following `dd`. But I think we shouldn’t do the before getting resolution on the spec issue.

PR #1743, opened 2020-08-05
Update spec data in parallel

### LINKED ISSUE #1741: Parallelize `bikeshed update`
Hi, I was reading through https://github.com/tabatkins/bikeshed/blob/e05aa73d9f0d0e0d3b51179dbc01f0dd5b8f30a5/bikeshed/update/manifest.py#L140 and it looks like it is updating a single file at a time.

Will it be a good idea to parallelize some of this? Or will it be rude to GitHub / hit some sort of API rate limit? Wondering if you've tried this before.

ngzhian:
Using queue and threads. The bulk of the logic is from the queue example
given in https://docs.python.org/3/library/queue.html#queue.Queue.join.
Usage of os.sched_getaffinity is explained
in https://docs.python.org/3/library/os.html?highlight=os#os.cpu_count.

I manually verified the correctness of this by comparing
bikeshed/spec-data/ of the current update with the one generated with
this patch applied.

$ time bikeshed update # current
real    2m17.526s
user    0m21.730s
sys     0m1.682s

$ time bikeshed update # with this patch, ran on 28 cores
real    0m13.869s
user    0m32.152s
sys     0m3.807s

We can probably speed this up more, maybe multiprocessing? But looks
good enough for me for now.

Fixed #1741.

### DIFF
--- a/bikeshed/update/manifest.py
+++ b/bikeshed/update/manifest.py
@@ -4,7 +4,9 @@
 import hashlib
 import io
 import os
+import queue
 import requests
+import threading
 from datetime import datetime
 
 from ..messages import *
@@ -143,28 +145,54 @@ def updateByManifest(path, dryRun=False):
         lastMsgTime = time.time()
         if newPaths:
             say("Updating {0} file{1}...", len(newPaths), "s" if len(newPaths) > 1 else "")
-        for i,filePath in enumerate(newPaths):
-            remotePath = ghPrefix + filePath
-            localPath = localizePath(path, filePath)
-            try:
-                newFile = requests.get(remotePath).text
-            except Exception as e:
-                warn("Couldn't download file '{0}'.\n{1}", remotePath, e)
-                return False
-            try:
-                dirPath = os.path.dirname(localPath)
-                if not os.path.exists(dirPath):
-                    os.makedirs(dirPath)
-                with io.open(localPath, 'w', encoding="utf-8") as fh:
-                    fh.write(newFile)
-            except Exception as e:
-                warn("Couldn't save file '{0}'.\n{1}", localPath, e)
-                return False
 
-            currFileTime = time.time()
-            if (currFileTime - lastMsgTime) >= messageDelta:
-                say("Updated {0}/{1}...", i+1, len(newPaths))
-                lastMsgTime = currFileTime
+        updateQueue = queue.Queue()
+        success = True
+        # This worker will download from remote, and update success if it fails.
+        def worker():
+            # Needed to update status and progress.
+            nonlocal success
+            nonlocal lastMsgTime
+            while True:
+                (i, filePath) = updateQueue.get()
+                remotePath = ghPrefix + filePath
+                localPath = localizePath(path, filePath)
+                try:
+                    newFile = requests.get(remotePath).text
+                except Exception as e:
+                    warn("Couldn't download file '{0}'.\n{1}", remotePath, e)
+                    success = False
+                try:
+                    dirPath = os.path.dirname(localPath)
+                    if not os.path.exists(dirPath):
+                        os.makedirs(dirPath)
+                    with io.open(localPath, 'w', encoding="utf-8") as fh:
+                        fh.write(newFile)
+                except Exception as e:
+                    warn("Couldn't save file '{0}'.\n{1}", localPath, e)
+                    success = False
+
+                currFileTime = time.time()
+                if (currFileTime - lastMsgTime) >= messageDelta:
+                    say("Updated {0}/{1}...", i+1, len(newPaths))
+                    lastMsgTime = currFileTime
+                updateQueue.task_done()
+
+        # Enqueue all the files to update. i is used for printing progress.
+        for i,filePath in enumerate(newPaths):
+            updateQueue.put((i, filePath))
+
+        # Create as many threads as we can.
+        for i in range(len(os.sched_getaffinity(0))):
+            t = threading.Thread(target=worker, daemon=True).start()
+
+        # Wait for all the requests to be done.
+        updateQueue.join()
+
+        # If any of the request fail, update has failed.
+        if not success:
+            return False
+
         try:
             with io.open(os.path.join(path, "manifest.txt"), 'w', encoding="utf-8") as fh:
                 fh.write("\n".join(remoteManifest))


bikeshed/update/manifest.py
```diff
@@ -143,28 +145,54 @@ def updateByManifest(path, dryRun=False):
         lastMsgTime = time.time()
         if newPaths:
             say("Updating {0} file{1}...", len(newPaths), "s" if len(newPaths) > 1 else "")
+
+        updateQueue = queue.Queue()
+        success = True
+        # This worker will download from remote, and update success if it fails.
+        def worker():
+            # Needed to update status and progress.
+            nonlocal success
```
jyasskin, line 154: Even Python has race conditions between threads, so results should be sent back using a queue also.

bikeshed/update/manifest.py
```diff
@@ -143,28 +145,54 @@ def updateByManifest(path, dryRun=False):
...
-
-            currFileTime = time.time()
-            if (currFileTime - lastMsgTime) >= messageDelta:
-                say("Updated {0}/{1}...", i+1, len(newPaths))
-                lastMsgTime = currFileTime
+            updateQueue.put((i, filePath))
+
+        # Create as many threads as we can.
+        for i in range(len(os.sched_getaffinity(0))):
+            t = threading.Thread(target=worker, daemon=True).start()
```
jyasskin, line 187: https://docs.aiohttp.org/en/stable/ suggests that we can avoid threads entirely by using a different fetching library.

bikeshed/update/manifest.py
ngzhian: Good point. There is only 1 result, success is initialized to `True`, and any threads that sees a failure will set it to `False`, and won't touch it otherwise.
Since we use a `.join()` below, any threads that have encountered a failure would have set `success` to `False`, and by the time we read `success` in the main thread, threads would have joined, so the value is consistent. Lmk if my understanding of this is wrong. Thanks!

bikeshed/update/manifest.py
ngzhian: Can you point to where in that page it mentions avoiding threads?
Also I'm hesitant to pull in an entire library just to make this parallel, but if @tabatkins is okay with it, I can do that!

tabatkins: These threads don't seem too complicated, so I'm fine with using them straight for now.

bikeshed/update/manifest.py
tabatkins: Yeah, since the only effect of `worker()` with the sync world is (a) possibly setting `success` to False, and (b) modifying a unique file, I think we're safe here.

jyasskin: And the race on `lastMsgTime` will _probably_ just result in writing the "Updated n/m" message twice. And, of course, the number of files the output claims to have already fetched can actually go down if HTTP responses come back out of order.

bikeshed/update/manifest.py
jyasskin: https://docs.aiohttp.org/en/stable/glossary.html#term-asyncio mentions that it's single-threaded, which reduces how many race conditions you need to handle.

bikeshed/update/manifest.py
```diff
@@ -143,28 +145,54 @@ def updateByManifest(path, dryRun=False):
...
-                return False
-
-            currFileTime = time.time()
-            if (currFileTime - lastMsgTime) >= messageDelta:
-                say("Updated {0}/{1}...", i+1, len(newPaths))
-                lastMsgTime = currFileTime
+            updateQueue.put((i, filePath))
+
+        # Create as many threads as we can.
+        for i in range(len(os.sched_getaffinity(0))):
```
jyasskin, line 186: Whether or not you use threads, the number of parallel HTTP fetches shouldn't be based on the number of usable CPUs. HTTP handling isn't CPU-bound, and Python threads run under a global lock that prevents them from using more than 1 net CPU anyway. (That's why `multiprocessing` exists, but again, because HTTP handling isn't CPU-bound, it won't help here.)

bikeshed/update/manifest.py
tabatkins: Ah yup the printing is indeed racy.

bikeshed/update/manifest.py
ngzhian: > Whether or not you use threads, the number of parallel HTTP fetches shouldn't be based on the number of usable CPUs.

Sure, this is merely a heuristic. If I have 10 cores, I should be able to run up to 10 threads that do blocking work. It's a quick way of running blocking network calls in parallel.

> Python threads run under a global lock that prevents them from using more than 1 net CPU anyway.

I don't know how the GIL interacts here, but from my testing, I see a speedup. If the GIL prevents these threads from being parallelized, why do we see a speedup?

> HTTP handling isn't CPU-bound, it won't help here

Yea, it's a network call. But in this case the network call blocks, you can imagine it's a CPU busy wait (for network response).

jyasskin: Sorry, I didn't write that clearly. The GIL prevents multiple Python threads from using CPU concurrently, but it gets released while a Python thread is waiting on an OS function, which is what happens while one of these HTTP fetches waits for the response to come back. That's *not* a busy-wait.

So, even with just 1 core, you can efficiently run lots of threads here to get that many HTTP fetches to be sent in parallel. They can't actually process results in parallel, but that's fine because processing results takes much less wall-time than waiting for the server to respond.

tabatkins:
Okay, I ended up rewriting this myself, relying on aiohttp (and aiofiles for the file creation). Thanks so much for the impetus to get this done, @ngzhian, and the pointer to aiohttp, @jyasskin!

On my laptop, updating went from ~3 files/second, to, uh, about **200 files/second**. I can regen the *entire* set of files now (about 1700) in less than 10 seconds. This is *astonishing*, I had no idea I was blocking so long on network like that, phew.

Anyway, this was great! Yay!

ngzhian:
Amazing, thanks Tab and Jeffrey!

ngzhian:
Heads up, I see that some of our builds are hanging, see https://travis-ci.org/github/WebAssembly/simd/builds/716629866 for example. Could be related to this.

ngzhian:
A local run of `bikeshed update` hangs at `Updating 44 files...`

ngzhian:
Seems like f539c4af782289625bc4591437f2f54176ab1d00 fixed it, we were pulling a broken alpha version (maybe)

PR #1767, opened 2020-09-23
Turn off more inline styles for WHATWG

annevk:
And also switch to a non-legacy syntax and align defaults between normal and RD.


annevk:
(I guess when landing this it can close https://github.com/whatwg/meta/issues/188.)

annevk:
@sideshowbarker was it intentional that -RD did not have `style-mdn-anno off, script-mdn-anno off`?

annevk:
I guess since Use Dfn Panels is off that doesn't matter. There are some other bits that seem out-of-sync though.

domenic:
Looking at https://dom.spec.whatwg.org/review-drafts/2019-06/ it seems that the mdn anno styles are done inline, so we probably shouldn't turn them off for review drafts yet. To fix that we'd need to move https://github.com/whatwg/whatwg.org/blob/7a2314f5c2ed2a29668b24d5f6bc7d94bd2f056a/resources.whatwg.org/standard-shared-with-dev.css#L96 into https://github.com/whatwg/whatwg.org/blob/master/resources.whatwg.org/spec.css

annevk:
So https://dom.spec.whatwg.org/review-drafts/2020-06/ is newer and doesn't include any MDN boxes. The shared-with-W3C version has a custom pipeline that's not under version control (yet) so I'm not sure that's a good reference.

tabatkins:
This looks good overall; feel free to merge whenever y'all want.

sideshowbarker:
> @sideshowbarker was it intentional that -RD did not have `style-mdn-anno off, script-mdn-anno off`?

Not intentional

PR #1771, opened 2020-09-24
Fold bikeshed-tests into bikeshed and run under GitHub Actions

foolip:


foolip:
I haven't been able to prove this working yet, because running the scripts run out of quota. I'll try triggering it again tomorrow to see if there's enough quota to run it once a day at least. If not this setup isn't going to work and a stored access token for a bot account will be needed instead.

foolip:
I just tried https://github.com/tabatkins/bikeshed/runs/1164481267?check_suite_focus=true and it still runs out of quota. The script will have to be rate limited or something.

foolip:
I implemented some throttling, which I'm actually sure is broken when the quota resets after about an hour, but in any case got past the GitHub API hammering in https://github.com/tabatkins/bikeshed/runs/1164954990?check_suite_focus=true. Then it fails when rebasing tests. @tabatkins can you take a look?

nschonni:
Took a slightly more basic approach in https://github.com/tabatkins/bikeshed/pull/1948.
Found you can also get the rate limit reset time, so you don't need to arbitrarily pause

foolip:
Cool, thanks @nschonni! @tabatkins can you review https://github.com/tabatkins/bikeshed/pull/1948?

tabatkins:
I'm integrating these two approaches by hand; the list of included/excluded things is important (rather than just running off the list of existing repos we know about), but I should be able to yank in some of the lessons from Nick's code too.

tabatkins:
Okay, after some light tweaking (notably, using Nick's code for determining the exact sleep necessary), I've merged the branch in manually. We'll see how it goes for a bit, then we can retire `bikeshed-tests` entirely.

nschonni:
I think the new throttle might not work as intended, because `get_rate_limit().search.remaining` was used for mine because I was using the code search API (which has a pretty tiny limit), but I don't think that's the API being called in the merged approach

PR #1772, opened 2020-09-25
Turn off style-dfn-panel for WHATWG

andreubotella:
See whatwg/whatwg.org#343, which moves those styles into WHATWG's standard.css stylesheet.

### DIFF
--- a/bikeshed/boilerplate/whatwg/defaults-RD.include
+++ b/bikeshed/boilerplate/whatwg/defaults-RD.include
@@ -4,7 +4,7 @@
   "Use Dfn Panels": "no",
   "Force Crossorigin": "yes",
   "No Editor": "yes",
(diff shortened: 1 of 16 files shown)


tabatkins:
I presume I should wait to merge this until the WHATWG pr is merged, so it doesnt' go without styling.

If Anne or Domenic want to merge them together, feel free.

annevk:
We should also change defaults-RD.include.

I can take care of merging this once all is in order.

andreubotella:
> We should also change defaults-RD.include.

I was under the impression that `"Use Dfn Panels": "no"` already disabled the styles, but I see I was wrong.

Btw, when I went to test this, `defaults-RD.include` was failing to parse because it was missing a comma. @tabatkins, you might want to add tests to catch this sort of stuff.

tabatkins:
> I was under the impression that "Use Dfn Panels": "no" already disabled the styles, but I see I was wrong.

Ah, this is because the use of dfn panels in the Index Of Externally-Defined Terms is *not* controlled by that metadata, but rather happens unconditionally.

I presume that should be similarly controlled?

annevk:
FWIW, even if Use Dfn Panels were to disable this, I'd rather continue to duplicate it in default-RD for ease of comparison between that and default.

PR #1774, opened 2020-09-29
Add missing continue statements in parseConditions

annevk:
Looks like this was missed in e632b349.

(The weird thing is that I cannot trigger this in a test. I guess the tests don't record fatal errors or warnings? That's kinda problematic for the WHATWG regression tests as our build system does break on those.)


tabatkins:
The tests purposely don't break on fatal errors, specifically so I can do illegal things and get that result recorded.

tabatkins:
Yup, been meaning to do that for a while, haven't gotten to it yet.

PR #1778, opened 2020-09-30
Make dfn panels readable in dark mode.

Kangz:


tabatkins:
Ah whoops, I just noticed the error myself and fixed it. Thanks for the PR anyway. ^_^

PR #1781, opened 2020-10-01
WHATWG: exlude LATESTRD on RDs

annevk:
Only the Living Standard should have this paragraph. See https://github.com/whatwg/whatwg.org/pull/344 for context.

We could also exclude this from LS-COMMIT and LS-PR, but it seems fine to me to include it there. (Note that adding LS to include-if would not work, as it's OR, not AND.)


domenic (approved):

annevk:
Oh my, the typo. Oh well.

tabatkins:
> (Note that adding LS to include-if would not work, as it's OR, not AND.)

You can get AND behavior by stacking up some `<if-wrapper>` elements, btw.

PR #1800, opened 2020-10-20
fixed immersive-web/status-ED for PP2017

himorin:
the same for 389ad75, added one to immersivewebwg/status-ED.include also.


tabatkins:
I'd have to verify with @fantasai, but I left immersive-web alone purposely because they are using the new patent policy, right?

fantasai:
No, it's WebRTC that was rechartered.

PR #1828, opened 2020-12-02
DAS WG: Use Patent Policy 2020

xfq:
The DAS WG operates under Patent Policy 2020 now: https://www.w3.org/2020/11/das-wg-charter.html#patentpolicy

/cc @reillyeon @anssiko @plehegar


anssiko:
LGTM

PR #1831, opened 2020-12-06
Support linking to steps.

agl:
Specifications often have algorithms with many (nested) ordered lists.
In those algorithms, steps often refer to other steps by number. But
those numbers currently have to be maintained manually and, when steps
are inserted, that can lead to a cascade of updated step number
references below. Worse, when two changes are in-flight to the same
algorithm, the step numbers are almost ensured to go wrong because
neither change need textually conflict with the other.

This change to bikeshed permits items in an ordered list to be given an
ID and referenced by name.

1. This is a traditional item.
1. {#myid} This is an item with an ID.
1. This item references [[#myid]].

The reference is turned into the string “step 2”. This also works for
nested steps where the reference will say something like “step 2.1.3”.

### DIFF
--- a/bikeshed/markdown/markdown.py
+++ b/bikeshed/markdown/markdown.py
@@ -8,14 +8,14 @@ from ..h import escapeAttr
 from ..messages import *
 
 
-def parse(lines, numSpacesForIndentation, features=None, opaqueElements=None, blockElements=None):
+def parse(lines, numSpacesForIndentation, features=None, opaqueElements=None, blockElements=None, itemNumContext=None):
     fromStrings = False
     if any(isinstance(l, str) for l in lines):
         fromStrings = True
         lines = [Line.Line(-1, l) for l in lines]
     lines = Line.rectify(lines)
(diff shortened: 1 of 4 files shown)


agl:
I have no idea what I'm doing, and sticking information in HTML attributes like this for later consumption is probably silly, but this issue comes up a bunch for WebAuthn / CTAP stuff and I figured I'd give this a shot.

@equalsJeffH to see whether I'm actually solving the right problem, or whether what we need is slightly different.

equalsJeffH:
> to see whether I'm actually solving the right problem, or whether what we need is slightly different

Thanks for doing this.  I'll have to figure out how to play with it once we ship the current on-deadline spec.

Nominally, what this supplies appears useful, especially if the resultant HTML is correct.   We've been up-to-this-point hacking this sort of thing into bikeshed source by using explict `<li id="foo"> markup, but bikeshed does weird things with that&mdash;such as altering the depth of the list item&mdash;more than occasionally such that the latter technique is not reliable, unfortunately. If this provides reliable results, then it is definitely useful.

ve7jtb:
Thanks,  but we won't delay RD02 to refactor quire yet if that is OK.

This will save a bunch of time and mistakes later.

PR #1841, opened 2020-12-22
Retrieve the correct metadata boilerplate if "Status" is specified.

### LINKED ISSUE #1818: WHATWG PR snapshots are wrong
It doesn't seem like anything in pr-preview changed so I suspect this might be a Bikeshed issue (though it's hard to find a suspicious change in the last month or so), but if you compare

* https://whatpr.org/dom/898.html
* https://whatpr.org/dom/909.html

you'll find that the first is correct and the second doesn't seem to have been generated in the correct way (e.g., it expects a COMMIT-SHA despite that only being for LS and not LS-PR). Anyone with ideas? I'm a bit at a loss.

cc @tobie @tabatkins @domenic

andreubotella:
In commit 9410d77, `config.retrieveBoilerplateFile` was changed to, among other things, use `doc.md.status` rather than `doc.md.rawStatus` as the default status specifier. However, when the `default` and `computed-metadata` boilerplates are retrieved, `doc.md.status` is set to `None` since the implicit metadata hasn't been computed yet, which causes the default boilerplate files to load rather than those for the specified status.

Fixes #1818 and fixes whatwg/whatwg.org#348.

### DIFF
--- a/bikeshed/config/retrieve.py
+++ b/bikeshed/config/retrieve.py
@@ -69,8 +69,11 @@ def retrieveBoilerplateFile(doc, name, group=None, status=None, error=True):
     # Filenames must be of the format NAME.include or NAME-STATUS.include
     if group is None and doc.md.group is not None:
         group = doc.md.group.lower()
-    if status is None and doc.md.status is not None:
-        status = doc.md.status
+    if status is None:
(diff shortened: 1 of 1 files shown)


tabatkins:
Nice catch, thanks!

PR #1844, opened 2021-01-06
chore: Enable Flake8 F841 rule

nschonni:
Giving the new setup a test since there are still 2 issues running flake
```
./bikeshed/issuelist.py:34:13: F841 local variable 'e' is assigned to but never used
./bikeshed/railroaddiagrams.py:188:9: F841 local variable 'token_patterns' is assigned to but never used
```

Edit: Added annotations, so the issues can be seen on the files tab https://github.com/tabatkins/bikeshed/pull/1844/files#file-bikeshed-issuelist-py-L34


tabatkins:
This is quite cool, thanks for putting this together. ^_^

nschonni:
No problem, the first one I think might be bad nesting of the `else`, but I wasn't sure about the `token_patterns` one. I'll see if I can figure something else so the CI goes green (after you sync the test results 😄 )

PR #1845, opened 2021-01-07
fix: remaining F841 errors

nschonni:


bikeshed/railroaddiagrams.py
```diff
@@ -184,27 +184,6 @@ def writeSvg(self, write):
             self.format()
         return DiagramItem.writeSvg(self, write)
 
-    def parseCSSGrammar(self, text):
```
nschonni, line 187: Couldn't find any references to this

bikeshed/issuelist.py
```diff
@@ -32,10 +32,8 @@ def printIssueList(infilename=None, outfilename=None):
                 infilename += suffix
                 break
             except Exception as e:
-                pass
-        else:
-            die("Couldn't read from the infile:\n  {0}", str(e))
-            return
+                die("Couldn't read from the infile:\n  {0}", str(e))
```
nschonni, line 35: The message was using the `e` from the exception out of scope, so I thought it should be hoisted

nschonni (commented):
My python is pretty basic, but added my reasoning below

bikeshed/issuelist.py
Ms2ger: I don't think this is correct, actually. It's pretty poor code, but (afaict) before the change, it would try three suffixes, and if they all threw, print the last exception; now it'll never get to the second run of the loop

nschonni: I'm good with changing this to something better 😄 
Maybe partially got back to the old except `pass`, but print them individually there, then `die` in the old place without trying to use the `e` out of scope
I'll let you send the PR though 😉

tabatkins: Aw dang indeed, this is real old code and I forgot how it worked.

nschonni: I opened #1852 with a partial backout

PR #1848, opened 2021-01-07
chore: Add Python 3.9 to CI matrix

nschonni:


nschonni:
Looks like Windows on 3.9 is having an issue with libxml. Going to see about the message `Using legacy 'setup.py install' for lxml, since package 'wheel' is not installed.`

requirements.txt
```diff
@@ -6,7 +6,7 @@ cssselect>=1.1.0,<1.2
 html5lib>=1.1,<1.2
 isodate>=0.6.0
 json_home_client>=1.1.1,<2
-lxml>=4.5,<4.6
+lxml>=4.5
```
nschonni, line 9: I wasn't sure if this is going to cause any issues with the Pipfile.lock, but it looks like dependabot also updated it now to 4.6.2, so maybe it should be >=4.6 now

nschonni:
@tabatkins gave this a rebase for the conflicts

PR #1852, opened 2021-01-08
fix: Try to open all files before final die

nschonni:


nschonni:
@Ms2ger does this make more sense?

bikeshed/issuelist.py
```diff
@@ -32,8 +32,11 @@ def printIssueList(infilename=None, outfilename=None):
                 infilename += suffix
                 break
             except Exception as e:
-                die("Couldn't read from the infile:\n  {0}", str(e))
-                return
+                failure("Couldn't read from the infile:\n  {0}", str(e))
```
tabatkins, line 35: It's not a failure here, this is meant to just be the "asking forgiveness is better than permission" thing that Python style encourages, grabbing the file if it exists rather than asking if it exists then grabbing it. So just revert it back to a `pass`.

nschonni: OK, should the outer `die` not use the `e` then? The original issue was the fact it was using that out of scope

tabatkins: Oh right. Hm, I guess it should hoist the e from one of the failures? Alternately, eh, it doesn't matter, just remove it.

nschonni: What if this is just a `say` instead of a `failure`. I looked in the messages.py and picked `failure` since it seemed bad based on the later `die`

tabatkins: No, it's not an error at all and shouldn't be announced. I'm just probing for possible files; the only error is if I don't find *anything*

nschonni: OK, removed the extra message and kept the `die` in the `else`, but without referencing the `e`

PR #1856, opened 2021-01-09
fix: Flake8 E401

nschonni:
Expanded glob imports at the same time for those files


tabatkins:
I suspect this is conflicting with <https://github.com/tabatkins/bikeshed/pull/1867>, mind rebasing?

PR #1850, opened 2021-01-07
fix: run pyupgrade --py37-plus on *.py

nschonni:
- ran `find . -path "*.py" -exec pyupgrade --py37-plus {} \;`
- Manually removed fonts.py call shebang to py 2.7
- Fixed "termref-for-" not passing refID to old .format
- Second pass seems to catch the previous io.open and fix default read

PR #1857, opened 2021-01-09
fix: Flake8 F523 unused format argument

nschonni:


nschonni:
@tabatkins cherry-picked the fix from #1850 to make it easier to land from that big review

PR #1861, opened 2021-01-09
fix: F522 '...'.format(...) has unused named argument(s)

nschonni:


nschonni:
Your extra cleanup is better 😄

PR #1865, opened 2021-01-09
fix: W605 invalid escape sequence

nschonni:


tests/algorithm001.html
```diff
@@ -508,7 +508,7 @@ <h3 class="no-num no-ref heading settled" id="index-defined-here"><span class="c
             }
         }
         function getVarName(el) {
-            return el.textContent.replace(/(\s| )+/, " ").trim();
+            return el.textContent.replace(/(\s|\xa0)+/, " ").trim();
```
nschonni, line 511: I don't think directly converting the character in the JS section was intended

tabatkins: Yeah, it should still *work*, but it's certainly not very clear when you're reading the generated script.

PR #1869, opened 2021-01-10
chore: Ignore BS functions defined in includes

nschonni:
Refrenced in publish.py and Spec.py.
Copied placeholders from boilerplate/bs-extensions.include

### DIFF
--- a/bikeshed/Spec.py
+++ b/bikeshed/Spec.py
@@ -289,7 +289,8 @@ class Spec:
                     "http://lists.w3.org"
                 ):
                     el.text = "https" + text[4:]
-            extensions.BSPrepTR(self)
+            # Loaded from .include files
+            extensions.BSPrepTR(self)  # pylint: disable=no-member
 
         return self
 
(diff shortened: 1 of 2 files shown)


tabatkins:
What prompted this? The code currently works fine, as far as I know.

nschonni:
It gets flagged in the IDE because it can't figure out that it's being defined (and overridden) in the `include` files

tabatkins:
Did you confirm that the correct versions of the functions get installed and called? This particular thing is still magic as far as I'm concerned.

nschonni:
Hmm, tried changing the CSS one to a junk value and the tests still passed, so I'll flip this to a draft till I can kick it a little more. I know the CI does tests this code because I got the tests to fail before when I remove the import in the `extensions.py` that brought in the `findAll` used in these files 😆

nschonni:
@tabatkins although the pylint rule isn't on globablly yet, I figured disabling it and adding a comment about the "magic-ness" of the loading made sense

PR #1871, opened 2021-01-11
chore: Move flake8 back to setup.cfg

nschonni:


tabatkins:
Good idea, but since I've made some additional changes, I did this manually in 3adee56b

(I'm done fiddling with flake8 settings now; it builds clean and every remaining ignore is intentional.)

nschonni:
Sounds good

PR #1872, opened 2021-01-11
fix: redefinition of 'line' in markdown.py

nschonni:
Partially reverted https://github.com/tabatkins/bikeshed/commit/283a7761e04c7bd7c8c8ad9b7c6786104912a5ba where it started to fail, then used VS Code to rename the second `l` symbol to `raw_line` since I couldn't figure out a better name


nschonni:
NVM, looks like you got it in a227bc87b36d0d16b191f3e5192272c0a29820f5

PR #1877, opened 2021-01-11
chore: Rerun black formatting

nschonni:


bikeshed/update/main.py
```diff
@@ -38,6 +38,7 @@ def update(
             say("Falling back to a manual update...")
             force = True
     if force:
+        # fmt: off
```
nschonni, line 41: I disabled this chunk because I think you want to keep this style of lining up the operators

PR #1884, opened 2021-01-12
fix: R1711: Useless return at end of function

nschonni:


nschonni:
Failure appears unrelated, so I spun up #1885 to fix the CI

PR #1887, opened 2021-01-13
fix: C1801: Do not use `len(SEQUENCE)` without comparison to determine if a sequence is empty (len-as-condition)

nschonni:


bikeshed/boilerplate.py
```diff
@@ -305,10 +305,10 @@ def addIndexSection(doc):
         container, E.h2({"class": "no-num no-ref", "id": safeID(doc, "index")}, "Index")
     )
 
-    if len(findAll(config.dfnElementsSelector, doc)):
+    if not findAll(config.dfnElementsSelector, doc):
         addIndexOfLocallyDefinedTerms(doc, container)
 
-    if len(list(doc.externalRefsUsed.keys())):
+    if not list(doc.externalRefsUsed.keys()):
```
nschonni, line 311: Feels like this could be unwrapped further to just `if not doc.externalRefsUsed:` but didn't want to play around too much

tabatkins: I suspect this was part of the py2->3 migration tool, where a bunch of things that switched from lists to generators got manually coerced to lists for safety.

bikeshed/boilerplate.py
```diff
@@ -305,10 +305,10 @@ def addIndexSection(doc):
         container, E.h2({"class": "no-num no-ref", "id": safeID(doc, "index")}, "Index")
     )
 
-    if len(findAll(config.dfnElementsSelector, doc)):
+    if not findAll(config.dfnElementsSelector, doc):
```
tabatkins, line 308: All of these conditions appear to be exactly opposite what they should be, yeah?

tabatkins (changes requested):

bikeshed/boilerplate.py
nschonni: Yeah, I went back and just added a value comparison instead of trying to do a (reversed) truthy

PR #1889, opened 2021-01-13
fix: W0104: Statement seems to have no effect

nschonni:


bikeshed/metadata.py
```diff
@@ -686,7 +686,7 @@ def parseRefStatus(key, val, lineNum):
     val = val.strip().lower()
     if val == "dated":
         # Legacy term that used to be allowed
-        val == "snapshot"
+        return "snapshot"
```
nschonni, line 689: My guess is that this was supposed to be `val = "snapshot` to coerce it so it would be returned in the next `if`, but the comparison seems to be a mistake

Issue #1897, opened 2021-01-19
Bikeshed update failing on Travis with a version conflict


rtoy:
WebAudio Travis CI clones bikeshed and does the normal pip3 install; bikeshed update.  This used to work fine.  Now bikeshed update is getting a version conflict.  See https://travis-ci.com/github/WebAudio/web-audio-api/builds/213692963

Perhaps related, when I do this on my local clone of bikeshed, `pip3 install --user -e .` produces the following error output.  I'm pretty sure this used to work.  `pip3 install -e .` also fails because I can't write to the installation directory /usr/local/lib/python3.8/dist-packages, which makes sense.  However, I'm pretty sure this used to work on Nov 30.

```
 Running setup.py develop for bikeshed
    ERROR: Command errored out with exit status 1:
     command: /usr/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/home/rtoy/work/github/tabatkins/bikeshed/setup.py'"'"'; __file__='"'"'/home/rtoy/work/github/tabatkins/bikeshed/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' develop --no-deps --user --prefix=
         cwd: /home/rtoy/work/github/tabatkins/bikeshed/
    Complete output (28 lines):
    running develop
    WARNING: The user site-packages directory is disabled.
    error: can't create or remove files in install directory
    
    The following error occurred while trying to add or remove files in the
    installation directory:
    
        [Errno 13] Permission denied: '/usr/local/lib/python3.8/dist-packages/test-easy-install-2024942.write-test'
    
    The installation directory you specified (via --install-dir, --prefix, or
    the distutils default setting) was:
    
        /usr/local/lib/python3.8/dist-packages/
    
    Perhaps your account does not have write access to this directory?  If the
    installation directory is a system-owned directory, you may need to sign in
    as the administrator or "root" account.  If you do not have administrative
    access to this machine, you may wish to choose a different installation
    directory, preferably one that is listed in your PYTHONPATH environment
    variable.
    
    For information on other options, you may wish to consult the
    documentation at:
    
      https://setuptools.readthedocs.io/en/latest/easy_install.html
    
    Please make the appropriate changes for your system and try again.
    
    ----------------------------------------
[...]

PR #1885, opened 2021-01-12
chore: pin dependencies and regenerate tests

nschonni:
Looks like the bug-fix release changed the way `const` is processed

PR #1898, opened 2021-01-19
chore: Remove pyproject.toml

nschonni:


nschonni:
@tabatkins I think removing this will address https://github.com/tabatkins/bikeshed/issues/1897 since it seems to have worked for @dbaron in https://github.com/tabatkins/bikeshed/pull/1885#issuecomment-761899740

PR #1899, opened 2021-01-20
fix: nested imports

nschonni:
- Hoist all but the previously ignored E402s
- Fix incorrect namespace for RawTokenFormatter

### DIFF
--- a/bikeshed/DefaultOrderedDict.py
+++ b/bikeshed/DefaultOrderedDict.py
@@ -1,3 +1,4 @@
+import copy
 from collections import OrderedDict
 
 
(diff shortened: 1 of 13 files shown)


tabatkins:
The nested imports in `cli.py` were intentional, as part of some perf hacking I did a few years back. (A few of the others might be too.) Since `cli.py` is the entry point for the entire program, I could shave off a measurable chunk of startup time by only importing the necessary modules that were needed for that keyword.

nschonni:
No problem if you want to keep it. Do you want me to pull out the RawTokenFormatter to another PR, since that is actually an error

tabatkins:
Yeah, go ahead, and we'll keep the rest of these for now; I'm generally fine with importing at function scope when used judiciously.

PR #1902, opened 2021-01-23
fix: no-member

nschonni:


bikeshed/InputSource.py
```diff
@@ -96,6 +96,11 @@ def cheaplyExists(self, relativePath) -> Optional[bool]:
         """
         return None
 
+    def __getattr__(self, name):
```
nschonni, line 99: Seems like a hack based on https://stackoverflow.com/a/60731663/455535

bikeshed/cli.py
```diff
@@ -620,7 +621,7 @@ def handleRefs(options, extras):
         exact=options.exact,
     )
     if constants.printMode == "json":
-        p(json.dumps(refs, indent=2, default=config.getjson))
+        p(json.dumps(refs, indent=2, default=getjson))
```
nschonni, line 624: `getjson` is a method off printjson, so adjusted with new import above. Not sure if this needs a test of some sort

bikeshed/stringEnum/StringEnum.py
```diff
@@ -24,3 +24,8 @@ def __contains__(self, val):
 
     def __getitem__(self, val):
         return self._vals[val]
+
+    def __getattr__(self, name):
```
nschonni, line 28: Same, hack as above, but makes more sense here since there are dynamic attributes set above

bikeshed/cli.py
```diff
@@ -579,7 +580,7 @@ def handleDebug(options, extras):
         doc = Spec(inputFilename=options.infile)
         doc.mdCommandLine = metadata.fromCommandLine(extras)
         doc.preprocess()
-        refs = doc.refs.refs[options.linkText] + doc.refs.refs[options.linkText + "\n"]
+        refs = doc.refs[options.linkText] + doc.refs[options.linkText + "\n"]
```
nschonni, line 583: I don't like this and I'm not sure if either way is correct. Could only guess that maybe the `getRef` might be a better thing, except didn't dig through that

nschonni (commented):
@tabatkins testing was passing for me locally except for the `RawTokenFormatter` that I have in a separate PR.
Didn't feel as confident with any of these fixes so have opened this as a draft

bikeshed/cli.py
tabatkins: Ooh yeah, that's some old code from back when the ReferenceManager class had a single `refs` member (now it's split across three members). This is obsoleted by the `bikeshed refs` command now, so we should just drop this whole `elif` and remove it from the command's options.

bikeshed/cli.py
tabatkins: This should be fine. It's clear that I don't use this codepath. ^_^

bikeshed/InputSource.py
tabatkins: Def a hack, but I'm okay with it if properly documented as such. One sec and I'll edit the commit...

tabatkins:
Okay, I'd be ready for merge except that CI is complaining about some unfixed `attribute-defined-outside-init` violations

nschonni:
It's unrelated, and re-suppressed in https://github.com/tabatkins/bikeshed/pull/1925
Not sure why it didn't fail on the original glob PR

PR #1905, opened 2021-01-23
chore: remove unused parameters in headings.py

nschonni:


tabatkins:
I'm inclined to leave this one; I prefer the more consistent "always pass the doc" function shape, especially since I've changed whether or not a particular function uses the doc over time before.

I don't see a change to a linting rule here - if we leave this, will it hurt any linting?

nschonni:
> I don't see a change to a linting rule here - if we leave this, will it hurt any linting?

It wasn't caught because of the globbing issue fixed in another PR, but I could suppress it at the file or function level when turning it back on

tabatkins:
Yeah, let's suppress at the file level

tabatkins:
Actually, I was misreading - I thought these were more general functions and I'd be annoyed having to edit several callsites if I needed to add a `doc` parameter later. They're not tho, they're just used for organization in this one file, so I'm fine with removed the unused params.

PR #1912, opened 2021-01-23
fix: pylint no-else-*

nschonni:


tabatkins:
I prefer using else/elif in these cases as an explicit semantic indicator that I'm dealing with equivalent code paths at the same level; I omit the `else` when the `if` is an exceptional test and the other code path is the "normal" path.  Can we drop an exception in pylint for this or something?

nschonni:
Yup, it will be ignored in https://github.com/tabatkins/bikeshed/pull/1907 and it can just be left disabled instead of re-enabling it.

PR #1928, opened 2021-01-28
fix: cell-var-from-loop

nschonni:
https://vald-phoenix.github.io/pylint-errors/plerr/errors/variables/W0640.html

### DIFF
--- a/bikeshed/boilerplate.py
+++ b/bikeshed/boilerplate.py
@@ -354,6 +354,21 @@ def addIndexOfLocallyDefinedTerms(doc, container):
     appendChild(container, indexHTML)
 
 
+def disambiguator(ref, types, specs):
+    disambInfo = []
+    if types is None or len(types) > 1:
+        disambInfo.append(ref.type)
+    if specs is None or len(specs) > 1:
+        disambInfo.append("in " + ref.spec)
+    if ref.for_:
+        try:
+            disambInfo.append("for {}".format(", ".join(x.strip() for x in ref.for_)))
+        except:
+            # todo: The TR version of Position triggers this
+            pass
+    return ", ".join(disambInfo)
+
+
 def addExplicitIndexes(doc):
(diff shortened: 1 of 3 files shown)


bikeshed/boilerplate.py
```diff
@@ -450,7 +449,9 @@ def disambiguator(ref):
         refsFromText = defaultdict(list)
         for ref in possibleRefs:
             for i, existingRef in enumerate(refsFromText[ref.text]):
-                if disambiguator(existingRef) != disambiguator(ref):
+                if disambiguator(existingRef, types, specs) != disambiguator(
+                    ref, types, specs
```
tabatkins, line 453: What's causing this sort of line-length protection to still happen? A different part of the formatting pipeline using a smaller limit than I specified in flake8?

nschonni: This was a warning/error from Black when I first pushed

bikeshed/update/updateCrossRefs.py
```diff
@@ -30,7 +34,7 @@ def update(path, dryRun=False):
         lastMsgTime = config.doEvery(
             s=5,
             lastTime=lastMsgTime,
-            action=lambda: say(f"Downloading data for spec {i}/{len(rawSpecData)}..."),
+            action=progress(i, len(rawSpecData)),
```
tabatkins, line 37: This is wrong - it's now eagerly evaluating the function (and thus passing `action=None`), rather than producing a lambda. I'll fix.

tabatkins: Ah, and I was wondering about this, because I know this code already works correctly; I'd have seen it if the closure was reusing bindings. Technically it is, but the lambda is immediately executed by the `doEvery()` and then thrown away, so it doesn't matter; this is a false-positive error.

That said, eh, it's fine to fix. (And also, I'm so glad JS got this *right* and establishes a fresh binding on every iteration of a for-of loop.)

PR #1933, opened 2021-01-28
chore: Format MD files with Markdownlint

nschonni:
If you look at the "smart diff" it doesn't actually change the end rendering, just removes some yellow squiglies if you use Markdownlint in your IDE


nschonni:
I didn't add a CI check for this since it doesn't seem like markdown is a big part of the repo for now

PR #1934, opened 2021-01-28
chore: Run lint before matrix

nschonni:
Not sure if this is a "good" optimization.
If you look at the runs, it now gates the lint in front of the full Python/OS matrix
image

Don't think it happens much, but since there is a 5 parallel macOS job limit, this will stop early if it can't pass the lint. Downside is that this potentially adds 2 minutes waiting for the lint to finish to a happy run

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build:
-
+    needs: lint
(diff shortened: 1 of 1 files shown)


tabatkins:
I do think this is good, actually. It's real annoying to wait for bots to cycle and then get stopped by a late linting failure. ^_^

PR #1938, opened 2021-01-28
fix: superfluous-parens

nschonni:
Kind of neutral on this one, since the readability may be better with them


tabatkins:
Yeah, there's one case here I'd be okay with changing, but the rest I'd prefer leaving in place, as the priority of `not` vs other things is not intuitively obvious to me (which is why I added the parens in the first place).

PR #1942, opened 2021-01-28
fix: few unused-argument

nschonni:


bikeshed/InputSource.py
```diff
@@ -75,7 +75,7 @@ def directory(self) -> str:
         """Suitable for passing to subprocess(cwd=)."""
         raise TypeError("{} instances don't have directories.".format(type(self)))
 
-    def relative(self, relativePath) -> Optional[InputSource]:
+    def relative(self, _) -> Optional[InputSource]:
```
nschonni, line 78: Maybe not the right fix. Comes from the original override signiture

bikeshed/datablocks.py
```diff
@@ -157,7 +156,7 @@ def transformDataBlocks(doc, lines):
     return newLines
 
 
-def cleanPrefix(lines, tabSize):
```
nschonni, line 160: Not sure if the `tabSize` concept got lost somewhere, since the function seems to indicate it might be relavant

tabatkins: Nah it's a leftover of the previous iteration of this code: https://github.com/tabatkins/bikeshed/commit/4ef7b2950cbfcdacc8f01e63a484b17cd52ba446

tabatkins:
All looks good except the shorthand ones - those aren't used, but the common contract for their methods has that signature, and I'd prefer to keep them all consistent. Let's put an exception on those lines.

Issue #1773, opened 2020-09-25
Release Notes


tabatkins:
Version 2.0.0
=============

* Feature: Added darkmode styles!

	This is the reason for the version bump, as if you were adding your own styles to your document, they probably aren't darkmode-aware and will look bad when the readers are in darkmode. If you're just using Bikeshed's own styles, tho, everything will work exactly as before.

* Feature: [Added new conditional-inclusion feature](https://tabatkins.github.io/bikeshed/#conditional), allowing you to remove sections of a spec (typically in boilerplate) based on the document's Status. If you have more things you want to base a conditional inclusion off of, let me know!

	(Currently, the processing order of this happens after all the 'datablocks' (`pre` with special class that triggers Bikeshed handling) happens, so even if those are inside of a conditional that gets excluded, they'll still get processed. But everything else happens after conditional stuff is removed.)

* Bugfix: [The rest of IDL's built-in types are now autolinked.](https://github.com/tabatkins/bikeshed/commit/428ace32b90c02fc39c872be53a3d916f51350de)

tabatkins:
Version 2.0.0
=============

* Feature: Added darkmode styles!

	This is the reason for the version bump, as if you were adding your own styles to your document, they probably aren't darkmode-aware and will look bad when the readers are in darkmode. If you're just using Bikeshed's own styles, tho, everything will work exactly as before.

* Feature: [Added new conditional-inclusion feature](https://tabatkins.github.io/bikeshed/#conditional), allowing you to remove sections of a spec (typically in boilerplate) based on the document's Status. If you have more things you want to base a conditional inclusion off of, let me know!

	(Currently, the processing order of this happens after all the 'datablocks' (`pre` with special class that triggers Bikeshed handling) happens, so even if those are inside of a conditional that gets excluded, they'll still get processed. But everything else happens after conditional stuff is removed.)

* Bugfix: [The rest of IDL's built-in types are now autolinked.](https://github.com/tabatkins/bikeshed/commit/428ace32b90c02fc39c872be53a3d916f51350de)

tabatkins:
Version 2.1.0
=============

[...]

PR #1946, opened 2021-01-30
Create python-publish.yml

nschonni:
Just a direct copy of https://github.com/nschonni/bikeshed/new/master?filename=.github%2Fworkflows%2Fpython-publish.yml&workflow_template=python-publish so it's probably the wrong folder (`dist`) and doesn't take into account your release script

This would probably mean switching to using Git tags + GitHub releases instead of https://github.com/tabatkins/bikeshed/issues/1773

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,0 +1,31 @@
+# This workflow will upload a Python Package using Twine when a release is created
+# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+
+name: Upload Python Package
+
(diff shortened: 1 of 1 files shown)


.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,31 @@
...
+  deploy:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
```
nschonni, line 20: Guessing maybe this should pin to 3.7 since that's the official version right now

PR #1947, opened 2021-01-31
fix: Skip metadata blank lines for specific blocks

nschonni:
Made empty list items and DDs in some cases

PR #1948, opened 2021-01-31
Add script and sync addional specs to tests

nschonni:
Based on a more basic approach from https://github.com/tabatkins/bikeshed/pull/1771.
- Pulls down newer versions of specs, and new BS files for the existing github repos
- Renames folders if the repo was renamed. May not handle the case where the repo is just "gone" rather than renamed
- Doesn't handle if the BS file was deleted on the remote. Those files will still stick around
- May need a concept for "ignores" like in 1771

Uncovered some other things like https://github.com/tabatkins/bikeshed/pull/1947 (also included here)


nschonni:
Ran the failing one locally again and I'm still getting a no-op after the original update
```
$ bikeshed test --rebase github/whatwg/quirks/review-drafts/2018-08.bs
1/1: Rebasing github/whatwg/quirks/review-drafts/2018-08.bs
```

tabatkins:
So just double-checking - this is *not* running under GH Actions, right? I'm not finding a file setting that up, at least.

And what does `dev.txt` do?

nschonni:
> So just double-checking - this is _not_ running under GH Actions, right? I'm not finding a file setting that up, at least.

No, just adds a manual syncing script here. The other PR has the more advanced Action + PR setup. This is just more of a manual `--rebase` equivalent to pull in new files.

> And what does `dev.txt` do?

It was just a "dev" version of `requirements.txt` for a pinned/working version of the dev dependencies (`pip3 install -r dev.txt` installs both the requirements.txt + dev.txt deps). If that makes sense to you, pinned pylint, flake8, and black versions

PR #1964, opened 2021-02-18
chore: Use GitHub Action task for PR creation

nschonni:
- Only run on the main repo, not forks
- Use peter-evans/create-pull-request for commit/PR work
- Don't run CI jobs on sync branch, just PR as it creates

Closes #1771


nschonni:
@tabatkins @foolip getting pretty close, but the current job failed during the commit creation, so I swapped to a GitHub Action task I've used before for nodejs/docker-node to do automated PRs
https://github.com/tabatkins/bikeshed/runs/1930321063?check_suite_focus=true

nschonni:
PS: It looks like when using the built-in `GITHUB_TOKEN` the rate limit is 1000 vs. if you change over to use a personal PAT, where it is 5000

foolip:
@tabatkins will you review this?

PR #1965, opened 2021-02-19
chore: Pylint/flake8 all in the CI

nschonni:
Just makes the root and new sync script get linted as well


nschonni:
@tabatkins gave this a rebase

PR #1976, opened 2021-03-05
fix: Rebase for single quote escape

nschonni:
Looks like the daily sync hit a snag today https://github.com/tabatkins/bikeshed/runs/2042842372?check_suite_focus=true#step:5:2259 and I think with the current behaviour, it won't try to rebase the files that haven't been touched

This is the changed behaviour from https://github.com/tabatkins/bikeshed/commit/803448c7fe43c1430ae24414bfab90d6003d7b48


tabatkins:
Ah, thanks, I was trying to get it rebased this morning but have some problematic test artifacts still sitting around, and then I had a long meeting so I didn't finish.

nschonni:
No problem, I added the second requirements file because it looked like it was running into a conflict between flake8 and pylint versions for some reason. Not sure if it needs a separate Dependabot job or not to bump them though

tabatkins:
Notably, this now conflicts with #1965, and I'm not sure what the desired change is.

PR #1977, opened 2021-03-06
fix: pointless-string-statement

nschonni:
Changed standalone string blocks to comments. Maybe they should be deleted instead


tabatkins:
Nah, I'm fine with these remaining strings, and fine with the pylint exception continuing.

PR #1999, opened 2021-03-25
Adjust boilerplates for Media WG to account for Notes

tidoust:
This updates the SOTD boilerplates for drafts that are intended to be published as WG Notes. Such specs will need to include a `Macro text: NOTE yes` macro definition in their Bikeshed metadata to trigger the right behavior.


tidoust:
Not sure why tests are failing. They don't seem to have liked the transition from http to https URLs. They seem correct though. @tabatkins what did I miss?

tabatkins:
Well, any change in the output will cause the tests to fail, because they no longer match the goldens. ^_^ I can take care of that, no worries.

More importantly, tho, why are you doing NOTE this way? You can already say `Status: NOTE` and key off of that in include-if/exclude-if instead.

tidoust:
> More importantly, tho, why are you doing NOTE this way? You can already say Status: NOTE and key off of that in include-if/exclude-if instead.

I may be missing something obvious, but what I would like in the end is not a NOTE, but rather:
1. in repositories, an ED of a NOTE
2. when published to /TR/, (a) a FPWD of a NOTE, then (b) a WD of a NOTE, and only after some time, (c) a NOTE.

My understanding is that `Status: NOTE` would give me 2.c. The pull request is more about getting 1, 2.a and 2.b.

tidoust:
> Well, any change in the output will cause the tests to fail, because they no longer match the goldens. ^_^ I can take care of that, no worries.

OK, I found the right test files and adjusted the URLs there.


The text for the notes should now be the right one. If there's a better mechanism to handle ED and WD of Notes, let me know! Otherwise, could this be merged so that we can enable automate publication for WebCodecs and accompanying registry specs?

nschonni:
@tidoust you can also run `bikeshed test --rebase` to regenerate the failing tests next time https://github.com/tabatkins/bikeshed/blob/main/CONTRIBUTING.md#running-tests

tabatkins:
Okay, I was still real confused about this, so I chatted with @frivoal about it:


> 16:58 TabAtkins: fantasai or florian: NOTE is a *status*, right? So you can't have ED vs WD vs whatever of NOTE?
>
> 16:58 TabAtkins: Or is it a track, and you can have various statuses on it?
>
> 16:58 florian: TabAtkins: more of a track.
>
> 16:58 TabAtkins: context: https://github.com/tabatkins/bikeshed/pull/1999#issuecomment-818860021
>
> 16:59 florian: TabAtkins: it is currently a bit of a mess, but it will be clean from the next edition of the process
>
> 16:59 TabAtkins: Hm, okay, I was under the assumption it was a status and designed some bits of Bikeshed accordingly.
>
> 16:59 TabAtkins: Small change, I guess, just add a NOTE-ED and NOTE-WD and whatever statuses.
>
> 17:00 florian: TabAtkins: for now: EDs are a separate things, not on any actual track, just the thing that you prepare. So Notes, have EDs, reck track things have EDs, (future) registry track things have EDs… anything can have an ED, and it's out of scope for the process
>
> 17:01 TabAtkins: Context up above is someone wanting to distinguish NOTE-track things from other things their WG produces (including EDs)
>
> 17:01 florian: other than that, Notes can be preceded by working drafts, but that's messy, as working drafts are supposed to be on the REC track and trigger the patent policy, which notes don't
>
> 17:01 TabAtkins: Right now they're proposing to handle it with a hack, using a special Text Macro to trigger the inclusion of some SotD text
>
> 17:01 florian: (also notes are used to retire things from the REC track, and that's odd, again for patent policy reasons as well as what-the-hell-is-this-thing reasons)
>
> 17:02 florian: from next year, there'll be a Note track, with (optional) Draft Notes, followed by Notes, which can (optionally) go through an AC Review to be elevated to W3C Statement
>
> 17:03 florian: (REC Track) Working Draft as a first stage of things meant to become Notes is an unfortunate reality at the moment, but we're in the process (aha!) of cleaning that up
>
> 17:05 florian: See these two PRs for how we're changing the process to deal with Notes:
>
> 17:05 florian: https://github.com/w3c/w3process/pull/488/commits
>
> 17:05 florian: https://github.com/w3c/w3process/pull/489
>
> 17:05 florian: (they're merged, but not yet voted into law by the AC)
>
> 17:07 florian: From a process PoV, EDs are completely out of scope. From a bikeshed point of view, I suspect it would be good to have REC-ED, Note-ED and Registry-ED as separate things, in order to get the status and other boilerplate to say relevant things for each track.
>
> 17:07 TabAtkins: Okay, that's the question I was about to ask.
>
> 17:07 TabAtkins: And then there's no such thing as "Note WD", right? That'd just be "Note"?
>
> 17:07 TabAtkins: or "Registry"?
>
> 17:08 florian: Currently, there's no dedicated Note WD, and if you want one, you use a a REC-track WD instead. Next year we're splitting the two, and there will be Note WDs (called Draft Notes).
>
> 17:09 TabAtkins: ...what's past Note WD, then?
>
> 17:09 TabAtkins: Is *that* just Note at that point?
>
> 17:09 florian: Note
>
> 17:09 TabAtkins: kk

So it looks like the correct path here is to introduce NOTE-ED and NOTE-WD statuses, and then use *that* to key the include/exclude of text. (And we should add similar statuses for REGISTRY.)

Sorry for the delay in figuring this out.

tidoust:
> So it looks like the correct path here is to introduce NOTE-ED and NOTE-WD statuses, and then use that to key the include/exclude of text. (And we should add similar statuses for REGISTRY.)

Yes, a NOTE-ED, NOTE-WD and NOTE-FPWD statuses are exactly what I need. And, later on when and if the process creates a registry track, REGISTRY-ED, REGISTRY-WD and REGISTRY-FPWD.

Context for this is the [WebCodecs Codec Registry](https://www.w3.org/TR/webcodecs-codec-registry/) and the [AVC (H.264) WebCodecs Registration](https://www.w3.org/TR/webcodecs-avc-codec-registration/) specifications. Both have been published as First Public Working Draft last week. Both are on the "note track" which, as @frivoal pointed out, is a clumsy mix of the Rec-track and not-the-Rec-track for the time being. First document would adopt the new REGISTRY track when it becomes available in the Process. Second one is intended to become a Note.

tabatkins:
All right, the new NOTE-* statuses patch looked good, so feel free to adapt this PR to use those and I'll happily merge.

tidoust:
Great! I adjusted the PR to use the new `NOTE-xx` statuses, merging all `status-XX.include` files into a single `status.include` file with the right `include-if` and `exclude-if` bits.

tabatkins:
Ah that's much better, thanks.

Issue #1895, opened 2021-01-16
Tests for multi-level specs, and <wpt-rest>


svgeesus:
Using the `<wpt-ignore>` tag results in the following warning

> WARNING: &lt;wpt-rest> is intended for debugging only. Move the tests to &lt;wpt>ents next to what they're testing.

Consider some specification CSS-foo, which has css-foo-3, css-foo-4, and css-foo-5, and assume each level adds features compared to the previous one. Remembering that WPT removed levels and will put all tests for all levels into wpt/css/css-foo/ then we have the following choices:

1. Use `<wpt-ignore>` in css-foo-3 to stop the tests from levels 4 and 5 being complained about; and again in css-foo-4 to stop it from complaining about tests from level 5; and maybe also in css-foo-5 because it is a delta spec or because you want to focus attention on just the new features
2. Use `<wpt>` but not `<wpt-ignore>`. Get used to ignoring a ton of warnings each time you build the spec locally
3. Don't use the wpt feature

None of these are ideal, so the actual benefit of the wpt tags is hard to realize for multi-level or multi-version specs.

tabatkins:
I'm not sure what you mean `<wpt-ignore>`. Are you referring to `<wpt-rest>`?

svgeesus:
OOps, yes

svgeesus:
It was even right there in the quoted warning.

tabatkins:
Unfortunately, I don't have a better suggestion. There simply isn't any WPT metadata indicating whether a given test is meant for a particular level. So yeah, we're stuck with a choice between "don't use wpt at all" and "get a persistent warning".  (Note that the "missing WPT tests" error message is much less obnoxious now, but I haven't updated pip with a version with the fix yet; I need to get on that. It's still a little longer than the wpt-rest message, at least.)

I think that a `<wpt-ignore>` element was requested in the past, where you could list tests that are technically under your prefix but that don't actually apply. Iirc the use-case it was requested for got solved a different way, but maybe that would help a little here? It would require regular maintenance (adding more entries to the old specs whenever tests were added for new features), but when it's up-to-date it would let you build cleanly without warnings. Would that help?

svgeesus:
Hmm, that might help. It would help if someone adds a new test so you get flagged; then you can either add it to your spec or add it to the exclude list.

 Needs input from more people before going that way.

frivoal:
[...]

Issue #1456, opened 2019-04-16
Change way to turn on wpt annotation


frivoal:
`WPT Display` can be set to `inline` in the metadata section. That's a very nice feature, but I think it's the wrong way to turn it on. I don't believe anybody keeps this turned on in the source that they commit. Adding and removing it from the document's source is an annoyance (and can lead to accidentally committing it).

Instead, having a command line argument for the same thing would seem to match usage a lot better.

Another possibility would be to always include the test info in the output, but hidden by default, with something in the document that can be clicked to make it appear.

tabatkins:
Command-line is already possible, by passing a `--md-WPT-Display=inline` flag (you can pass any metadata that way) <https://tabatkins.github.io/bikeshed/#overriding-metadata>

But your second suggestion of an option that hides them by default, with a switch to make them turn on, is a good one. Happy to add that.

Issue #1116, opened 2017-10-20
Inline WPT test information into specs


tabatkins:
# Background

Talking with fantasai, a major hurdle she has with dealing with testing is figuring out what is un/undertested. WPT has a view that at least sorts tests by *ToC section*, but that's not enough - you still have to manually read all of the tests and reverse them to testable statements in the spec yourself, annotating your personal copy, and *then* look at what you have and what wasn't touched.  Better would be to have test information inline in the spec, where after each paragraph (or even sentence!) you list the tests associated with that bit of text. This gives you an immediate view in the source of what's untested, and makes it easy to keep the tests in the right place when you move stuff around in the spec.

## Benefits / Drawbacks

* This is slightly more continuing work for the spec author, as they have to keep the tests up-to-date in their spec. It's substantially less work for the test-writer/maintainer, as they have an immediate view of what things are tests and what aren't. (In many cases these two people are the same, so it's a net win, I think.)
* Technically duplicates existing data somewhat, as tests currently link to things they're testing. However, most such links are very coarse-grained; useful enough for reviewing the test, but of minimal usefulness for reviewing the test _suite_. They can also drift out of sync if the spec moves text around, with no mechanism for keeping them in sync; a longer-term goal here could be to replace the link-to-spec with automatically-extracted information from the spec itself, displayed in the test harness.
* This produces "noise" in the spec, interrupting the flow of thought in the source. It's useful noise, and often won't be too much; I should investigate common editor's code-folding behavior and make sure the planned syntax is compatible with it.
* WPT and the spec can now be kept in perfect sync, and with minimal adjustment needed for changes on either side. Moving a WPT folder just means updating the single metadata line in the spec declaring the path prefix (and in the meantime, the spec will throw a thousand errors, or maybe I'll detect that specially and throw a single one).  Adding new tests to WPT will alert the spec author next time they build the spec, making it easy to review and insert them where needed.
* There's an existing large backlog of specs with tests in WPT but no inline annotations, requiring a large update to add them all. On the other hand, it's a one-time cost, with clear benefits, and can be done relatively mindlessly by anyone in a PR, not just the spec author.

## Why Hasn't This Been Done Before?

[...]

PR #2010, opened 2021-04-03
Miscellaneous improvements for the WPT element

frivoal:
Here's a loosely related set of tweaks, fixes, and new features to how the wpt elements work. I suppose I could send that as a bunch of separate PRs, but that seems like unnecessary overhead.
Some issues covered by this:
* https://github.com/tabatkins/bikeshed/issues/1456
* https://github.com/tabatkins/bikeshed/issues/1895
* some aspects of https://github.com/tabatkins/bikeshed/issues/1116#issuecomment-395851865

Not sure if this is easier to review commit by commit or not, but each commit should be clean, so you can if you want to, and I suspect the commit messages will give a bit of context about why I'm doing what I'm doing. One of the commits (https://github.com/tabatkins/bikeshed/commit/4e5384211602bd92b8bac9e435e76fc17608c96e) changes the indentation of a large block of code, so at least that one is probably easier to review with `--ignore-space-change`

Some of these choices are probably subjective, but I'm pretty happy with the result. Let me know if anything isn't to your liking.

I think that with these in place, I'd be happy to turn `WPT Display` on by default on all my EDs:  it would encourage readers in general (and implementers in particular) to look up the results of tests as they read various sections of the spec to see how well implementations are doing (and possibly file/fix bugs). Also, it would make it easier to notice when some part of the document doesn't have tests, and encourage writing them. (As for turning it on on TR, we'd probably have to sync up with pubrules and tr-design, but one thing at a time).

Maybe we could add `"WPT Display": "closed"` to `boilerplate/csswg/defaults-ED.include`. Besides giving us a good bunch of annotated EDs, that'll nicely populate the tests as well.


frivoal:
Also, as a possible alternative to the right floats for the closed state, another variant I quite like is with inline-blocks. You can try it out by applying this after the latest commit.
```diff
diff --git a/bikeshed/wpt/wptElement.py b/bikeshed/wpt/wptElement.py
index 9c6b614a9..76295f62b 100644
--- a/bikeshed/wpt/wptElement.py
+++ b/bikeshed/wpt/wptElement.py
@@ -323,18 +323,14 @@ wptStyle = """
     color: var(--wpt-border);
 }
 .wpt-tests-block:not([open]){
-    border-left: none;
-    background: none;
     margin: 0;
-    padding: 0;
-    border-top: 1px dotted var(--wpt-border);
+    padding-bottom: 0;
+    border: none;
     opacity: 0.5;
+    display: inline-block;
 }
 .wpt-tests-block:not([open]) summary{
-    background: var(--wpt-bg);
     padding: 0;
-    float: right;
-    max-width: max-content;
     font-size: 0.8em;
     line-height: 1;
 }
```

frivoal:
Never mind the previous comment, I've changed my mind. I had proposed some tentative changes to how things could be styled, but have since found a better way and included that in the pull request.

frivoal:
> Maybe we could add "WPT Display": "closed" to boilerplate/csswg/defaults-ED.include.

@fantasai had a look at the output and liked it, so I've tentatively included this in the PR as well.

tabatkins:
All right, finally had a chance to look at this, and it looks *great*. I'll probably actually log a breaking change to make "closed" the default, so people can prep by turning it off if they want.

PR #2014, opened 2021-04-05
Bio fixes

### LINKED ISSUE #2003: [[SHORTNAME inline]] doesn't do what is advertised
Bikeshed lets me do this, which is very nice:

> Browser must foo the bar, as defined in [[shortname#relevant-section]].

The link will have "spec title § section number section name" as it's text, and point to the right section in that spec. All good.

Now, if I want to follow the same pattern, but point to a the entire spec, according to [the documentation](https://tabatkins.github.io/bikeshed/#biblio-mods), this should do it:

> Browser must foo the bar, as defined in [[shortname inline]].

Unfortunately, it doesn't work. The link text is the spec's title, as expected, but the link takes you to the bibliography just as [[shortname]] would, which isn't what the documentation says (it claims "the link goes straight to that document, rather than going to the bibliography first").

If this is not intended, there's a bug. If it is intended, the documentation is wrong. But in that case, I'd still want a feature that does what the documentation talks about.

frivoal:
Fixes #2003, and a couple more bugs I found while fixing it.

### DIFF
--- a/bikeshed/unsortedJunk.py
+++ b/bikeshed/unsortedJunk.py
@@ -936,7 +936,10 @@ def processBiblioLinks(doc):
 
         biblioDisplay = el.get("data-biblio-display", doc.md.defaultBiblioDisplay)
         if biblioDisplay == "inline":
-            replaceContents(el, ref.title)
+            if (
+                el.text == "[" + linkText + "]"
+            ):  # False if it's already been replaced by an author supplied text using the [[FOOBAR inline|custom text]] syntax.
+                replaceContents(el, ref.title)
             if ref.url is not None:
                 el.set("href", ref.url)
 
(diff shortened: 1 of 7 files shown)


bikeshed/unsortedJunk.py
```diff
@@ -936,7 +936,12 @@ def processBiblioLinks(doc):
 
         biblioDisplay = el.get("data-biblio-display", doc.md.defaultBiblioDisplay)
         if biblioDisplay == "inline":
-            replaceContents(el, ref.title)
+            if (
+                el.text == "[" + linkText + "]"
```
nschonni, line 940:
```suggestion
                el.text == f"[{linkText}]"
```

PR #2040, opened 2021-04-28
Fix "This version" link for NOTE-WD status

tidoust:
This was missed when the NOTE-WD status was introduced. The prefix part in the "This Version" link needs to remain "WD" in that case, and not "NOTE-WD".

Also, it turns out that pubrules does not require the sentence "This document is informative only." in the SoTD of documents that will become Notes. Adjusting the boilerplate for Media WG documents accordingly.


tidoust:
@tabatkins Anything blocking this?
(I'm trying to automate publication of these note drafts to /TR and this makes publication fail)

PR #2057, opened 2021-05-12
Fix link to IPR page for GPU for the Web WG

tidoust:
Number was that of the Media WG. Took the opportunity to switch to the new URL pattern for these URLs.

Also add required "First Public" when document is published as FPWD.

### DIFF
--- a/bikeshed/boilerplate/gpuwg/status.include
+++ b/bikeshed/boilerplate/gpuwg/status.include
@@ -31,7 +31,11 @@
   This document is a <strong>First Public Working Draft</strong>.
 </p>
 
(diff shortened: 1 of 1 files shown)


tabatkins:
All right, merged, and took the opportunity to push version 2.4.1 to pypi. ^_^

tidoust:
Many thanks! Publication of WD using spec-prod now works like a charm for would-be notes in the Media WG.

(It looks like I forgot to update yet another part for the boilerplate of the GPU for the Web WG though, as the group already switched to the new Patent Policy. I'll create another PR for that)

PR #2062, opened 2021-05-20
fixed IWWG header boilerplate with recent? change, like status to sotd

himorin:
fixes to match with pubrules rules.
seems forgot to push this when I've worked on manual publishing of IWWG spec to /TR/ last time... (was kept in local stash)

@AdaRoseCannon , do we want to include our favicon lines here, instead of having in each spec .bs??


tabatkins:
Yeah, any boilerplate that you put in each spec should just be put in your boilerplates files instead. ^_^

himorin:
closing, this moved to bikeshed-boilerplate repo

PR #2076, opened 2021-06-04
Use datatracker.ietf.org URLs for RFC 2119

xfq:
tools.ietf.org is no longer to be relied on per https://mailarchive.ietf.org/arch/msg/tools-discuss/oYrAxb3KayPzZ4SNB1DVZTDPPNo/


hober (approved):

tabatkins:
Sigh, unnecessary linkrot.

I'll merge in a bit when I can rebase the tests.

PR #2086, opened 2021-06-16
DAS WG: Adjust the order of the text

xfq:
According to https://github.com/w3c/specberus/blob/2d79e706121ef49138927f9212a52ca7522568f2/lib/rules/sotd/stability.js#L19-L22 , the order of this text seems to be wrong.


tabatkins:
Merged, and I cut version 2.4.4 for you to include just this commit. ^_^

xfq:
Sorry, but I need to adjust this text further to pass Pubrules. I will send a new PR.

PR #2111, opened 2021-07-15
Add support and boilerplate for the WebExtensions Community Group (WECG).

hober:
It's been a while since I added a group; please make sure I edited all the right places, @tabatkins!


hober:
Is `wecg` okay for the group's bikeshed shortname, @dotproto @xeenon? or should it be `webextensions`?

dotproto:
Between the two I prefer `wecg`. At this point we don't know where our work will take us and `webextensions` may be confusing if we end up establishing a WebExtensions Working Group in the future.

PR #2117, opened 2021-07-23
Allow an introduction in issues lists

frivoal:
Here's an example of a file that uses that: https://github.com/w3c/w3process/blob/main/issues-20210603.txt


frivoal:
Never mind, look at this file instead: https://github.com/w3c/w3process/blob/main/issues-20210603.txt

frivoal:
Nevermind the old file. Look here instead: https://github.com/w3c/w3process/blob/main/issues-20210603.txt

PR #2130, opened 2021-08-28
CSSWG: fix FPWD boilerplate

xfq:
See https://www.w3.org/TR/2021/WD-css-cascade-5-20210119/#status for an example.


xfq:
Why is this closed? I'm not aware of an update to the Pubrules, and the previous version was not correct per Pubrules.

See https://www.w3.org/TR/2022/WD-vc-data-integrity-20221110/#sotd for a recent FPWD publication.

tabatkins:
Ah, sorry, the files it edited were moved to the bikeshed-boilerplate project over a year ago, so I figured it must have been fixed over there.

xfq:
I see. Thank you.

PR #2133, opened 2021-09-03
Add link for Applies to in propdef tables. Fixes #1992

### LINKED ISSUE #1992: Link propdef "Applies to" to definition in cascade
https://www.w3.org/TR/css-cascade/#applies-to

fantasai:


tabatkins:
We'll need to regen the tests as well

PR #2182, opened 2021-11-27
Autogenerate requirements.txt

### LINKED ISSUE #2181: Installation by `pip install bikeshed` still uses outdated dependencies
It installs Pillow 8.1.1 which does not support Python 3.10. It seems Pipfile has 8.3.2 instead, and the internet says Pipfile replaces requirements.txt, so why do both exist here?

saschanaz:
Dependabot will detect and update autogenerated requirements.txt. (https://github.com/dependabot/feedback/issues/435#issuecomment-483259640, https://github.com/Taxel/PlexTraktSync/pull/533/files) It adds more requirements because those are what Pipfile.lock has.

(I think it can ultimately be removed, since requirements.txt is not required if `pipenv` is used.)

Fixes #2181


tabatkins:
Phew, finally took the time to understand what was happening here and merged it in. Thanks for the work, @saschanaz !

PR #2191, opened 2021-12-06
Add a new "direct" biblio display style

### LINKED ISSUE #2032: Add an option to allow inline linking to use shortnames
Hi @tabatkins!!!!

Typically, when I insert a reference, e.g. `[[P1000]]`, the link text is the shortname of the reference, e.g. [[P1000]](https://wg21.link/P1000).

However, if I use the awesome `[[FOO inline]]` or `Default Biblio Display: inline` feature from #1641, then the link text is the full name of the reference, NOT the shortname, e.g. [C++ IS Schedule](https://wg21.link/P1000) instead of [[P1000]](https://wg21.link/P1000).

I'd like to be able to do inline links to biblio references which use the shortname as link text, not the full name.

brycelelbach:
Add a new biblio display style, "direct", which displays references by their shortname and links to the referenced document. Closes #2032.

### DIFF
--- a/bikeshed/shorthands/oldShorthands.py
+++ b/bikeshed/shorthands/oldShorthands.py
@@ -345,7 +345,7 @@ biblioRe = re.compile(
                         \[\[
                         (!)?
                         ([\w.+-]+)
-                        (\s+(?:current|snapshot|inline|index|obsolete)\s*)*
+                        (\s+(?:current|snapshot|inline|index|direct|obsolete)\s*)*
                         (?:\|([^\]]+))?
                         \]\]""",
     re.X,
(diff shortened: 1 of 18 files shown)


bikeshed/shorthands/oldShorthands.py
```diff
@@ -380,10 +380,17 @@ def biblioReplacer(match):
 
     displayInline = "inline" in modifiers
     displayIndex = "index" in modifiers
-    if displayInline and displayIndex:
-        die(f"Biblio shorthand {match.group(0)} contains *both* 'inline' and 'index', please pick one.")
-    elif displayInline or displayIndex:
-        attrs["data-biblio-display"] = "inline" if displayInline else "index"
+    displayDirect = "direct" in modifiers
+    if (displayInline + displayIndex + displayDirect) > 1:
```
tabatkins, line 384: I hate that Python bool is technically a subclass of int, but I suppose it has its uses...

docs/index.bs
```diff
@@ -1070,10 +1070,11 @@ There are several additional optional keys:
...
 		for bibliography and autolinks entries that have both "current" and "snapshot" URLs.
 		(You can also specify this per-biblio entry or per-autolink.)
-	* <dfn>Default Biblio Display</dfn> takes the values "index" (default) or "inline",
+	* <dfn>Default Biblio Display</dfn> takes the values "index" (default), "inline", or "direct"
 		and selects whether biblio autolinks default to
-		displaying as their shortname and linking to the bibliography index,
-		or displaying as their title and linking straight to the referenced document.
+		displaying as their shortname and linking to the bibliography index ("index"),
+		displaying as their title and linking straight to the referenced document ("inline"), or
+    displaying as their shortname and linking to the referenced document ("direct").
```
tabatkins, line 1077: your indent is broken

docs/index.bs
```diff
@@ -2928,13 +2929,14 @@ after the biblio name.
 * By default, biblio autolinks are formatted like you typed, just with a single set of square brackets:
 	`[[FOO]]` will produce a link with the text "\[FOO]",
 	which links to the bibliography.
+  You can make them display "direct", where they link straight to the document,
```
tabatkins, line 2932: broken indent

tabatkins (changes requested):
The feature looks good. Only problem is the busted indents in the docs, which'll break the formatting when I actually process them.

When you fix those, would you mind running Bikeshed over it and including the results in the PR, too?

brycelelbach:
D'oh. Looks like I missed the fact that the docs use tabs! I was wondering if I should regenerate them.

I've fixed the indents and added regenerated docs to the PR. I also rebased on `main` (and regenerated the tests and docs afterwards just to be safe).

brycelelbach:
Thanks!

PR #2227, opened 2022-02-10
Remove unnecessary type attribute from <style> and <script>

zcorpan:
The attributes are obsolete and are generating warnings in validator.w3.org/nu as per https://html.spec.whatwg.org/multipage/obsolete.html#warnings-for-obsolete-but-conforming-features

This is similar to https://github.com/tabatkins/bikeshed-boilerplate/pull/16


tabatkins:
Tho, since the built-in boilerplates are dramatically out of date at this point *anyway*, just fixing this issue will be putting a bandaid on a bullet wound. So nm, I'll just go ahead and close; I've been wanting to do a `readonly` update for a while, it just requires a lot of boring verification that everything's still as expected.

PR #2232, opened 2022-02-23
Add support for 'permission' dfn type

dontcallmedom:
As discussed in https://lists.w3.org/Archives/Public/spec-prod/2022JanMar/0008.html


marcoscaceres:
sorry, how does one actually use it? The docs are not very clear on how one sets a definition type.

tabatkins:
Like any other definition type, as a boolean attribute on the dfn: `<dfn permission>foo</dfn>`. (Or as a value in the `dfn-type` attribute on an ancestor.)  Similar with links: `<a permission>`.

marcoscaceres:
Ah, thanks!

PR #2269, opened 2022-04-22
Allow adding `assert` to a div tag.

### LINKED ISSUE #2268: Wrapping assert tags around block elements?
In the WGSL spec we would like to start using the `assert` capabilities of bikeshed to annotate the spec. We started adding asserts in https://github.com/gpuweb/gpuweb/pull/2783 but ran into the issue that a lot of what we'd like to assert on are bullet point lists or tables.

So, what we'd like to do is:
```
<assert>
* requirement 1
* requirement 2
* requirement 3
</assert>
```

While this generates in bikeshed, it fails the HTML validation because the `<ul>` element ends up embedded inside a `<span>`.

Is this something that could be handled in bikeshed? Possibly by doing something like generating the assert from the contents of the list, but not wrapping the list in the `<span>` tag in the generated content? Or something like having parent/child assert spans so the list elements have their own span ids, but they link to a parent span that covers all of them?

dj2:
This PR adds the ability to add `assert` to a div tag to generate the
unique assert hash for the contents of the div and then attach an
`id=assert-<hash>` to the div. This allows using the assert
functionality with larger block based pieces of spec.

Fixes #2268

### DIFF
--- a/bikeshed/unsortedJunk.py
+++ b/bikeshed/unsortedJunk.py
@@ -1215,7 +1215,7 @@ def cleanupHTML(doc: "t.SpecType"):
             el.tag = "span"
             el.set("id", h.safeID(doc, "assert-" + h.hashContents(el)))
         # For any <div assert> elements, add a unique ID based on its contents.
-        if el.tag == "div" and "assert" in el.attrib:
+        if el.tag == "div" and el.get("assert") is not None:
             h.removeAttr(el, "assert")
             el.set("id", h.safeID(doc, "assert-" + h.hashContents(el)))
 
--- a/bikeshed/unsortedJunk.py
+++ b/bikeshed/unsortedJunk.py
@@ -1207,11 +1207,16 @@ def cleanupHTML(doc: "t.SpecType"):
             el.tag = "span"
             h.addClass(el, "css")
 
+        # Assert IDs are used to tag arbitrary sections with an ID so you can point tests at it.
+        # (The ID will be guaranteed stable across publications, but guaranteed to change when the text changes.)
+        #
         # Transform the <assert> fake tag into a span with a unique ID based on its contents.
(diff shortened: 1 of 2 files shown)


dneto0 (commented):
+1
This seems like a really nice solution.
As WGSL editor, I would really like this.

dj2:
Ping @tabatkins, does this seem like a useful addition?

bikeshed/unsortedJunk.py
```diff
@@ -1207,12 +1207,17 @@ def cleanupHTML(doc: "t.SpecType"):
...
+        # (The ID will be guaranteed stable across publications, but guaranteed to change when the text changes.)
+        #
         # Transform the <assert> fake tag into a span with a unique ID based on its contents.
-        # This is just used to tag arbitrary sections with an ID so you can point tests at it.
-        # (And the ID will be guaranteed stable across publications, but guaranteed to change when the text changes.)
         if el.tag == "assert":
             el.tag = "span"
             el.set("id", h.safeID(doc, "assert-" + h.hashContents(el)))
+        # For any <div assert> elements, add a unique ID based on its contents.
+        if el.tag == "div" and "assert" in el.attrib:
```
tabatkins, line 1218: Use `el.get("assert") is not None` here, to match style elsewhere.

tabatkins (changes requested):
One minor nit in the code, then r+. Ideally you could update the docs as well to talk about this?

bikeshed/unsortedJunk.py
dj2: done

dj2:
Moved the doc into an Assertion section. Please take a look and let me know what you think.

dj2:
ping @tabatkins

tabatkins:
Yup, just realized on Tuesday I needed to let the bots cycle. Looks good to go now. :+1:

PR #2273, opened 2022-05-02
Support building multi-file specs from tar files

kainino0x:
Allows tar files containing multiple source files to be passed to `bikeshed spec`. The tar file must contain an `index.bs`. Cross-file includes (`pre class=include`) are supported between files inside the archive, so a multi-file spec can be archived and then built from a single tar file.

This makes it easy to do multi-file builds using the Bikeshed API (http://api.csswg.org/bikeshed/). Bikeshed operates on the tar file in memory, avoiding extracting arbitrary files to the local filesystem. It disallows compression to avoid any surprises (zip bombs).

For an input `myspec.tar`, the default output is `myspec.html`.

Usage:
- Output to a file:
  `bikeshed spec myspec.tar myspec.html`
  `bikeshed spec myspec.tar`
- Output to stdout:
  `bikeshed spec myspec.tar -`


bikeshed/InputSource.py
```diff
@@ -157,26 +158,46 @@ def relative(self, relativePath) -> UrlInputSource:
...
-                datetime.datetime.fromtimestamp(os.path.getmtime(self.sourceName)).date(),
-            )
+        if self.tarFile:
+            try:
+                mtime = self.tarFile.getmember(self.sourceName).get_info()["mtime"]
+            except KeyError as e:
+                raise FileNotFoundError(errno.ENOENT, "Not found inside tar file", e.args[0]) from e
+            with self.tarFile.extractfile(self.sourceName) as f:
+                # Decode the `bytes` to a `str`. (extractfile can't read as text.)
+                file_contents = f.read().decode(encoding="utf-8").splitlines(keepends=True)
```
tabatkins, line 194: Rather than doing a test poke and *then* decoding the file, just put the file read inside the `try` and catch whatever error `extractfile` throws when the file doesn't exist.

tabatkins:
Ooooh, pretty cool overall. Some significant questions/nits before I merge, tho.

1. Is there a particular reason you merged this functionality into FileInputSource rather than making a new TarInputSource? It looks like the two don't actually share much code, and for the methods you didn't add stuff to, it's not clear to me that they're correct.
2. The tarFile shouldn't be a global constant; it's particular to a given Spec file and should be a property of that object. (Or rather, probably just encapsulated in the `InputSource` object?)
3. Rather than passing the input path in the --tar-file argument and ignoring the infile argument, I think `--tar-file` (or just `--tar`?) should be a boolean, and we can still use the infile argument for the path.
4. The outfile argument still makes sense, and I think the default should be to write to a `[same name].html` file in the same directory as the tarfile, like we currently do for plain BS files.

kainino0x:
Thanks for the review! I'll look into revisions soon. I think your suggestions all make sense.

> * Rather than passing the input path in the --tar-file argument and ignoring the infile argument, I think `--tar-file` (or just `--tar`?) should be a boolean, and we can still use the infile argument for the path.

I did this because if you had a `.bs` file with a name other than `index.bs` or `Overview.bs` then this would let you specify the filename inside the tar. But that doesn't matter for us (we use `index.bs`) and you can always work around it by renaming your toplevel file. So I'm happy to simplify this.

kainino0x:
2/4: Done.

> 1\. Is there a particular reason you merged this functionality into FileInputSource rather than making a new TarInputSource? It looks like the two don't actually share much code, and for the methods you didn't add stuff to, it's not clear to me that they're correct.

TBH because my screen was too small when I was doing this and I didn't see that FileInputSource was a small class I could reimplement ^_^

> 3\. Rather than passing the input path in the --tar-file argument and ignoring the infile argument, I think `--tar-file` (or just `--tar`?) should be a boolean, and we can still use the infile argument for the path.

I did roughly this but without adding an argument at all - it just detects `.tar` in the filename.
Do you think this is OK? It seems easier for bikeshed-web (assuming it preserves the filename, which I'm not sure about) because then hopefully it won't need any updates.

kainino0x:
> 2/4: Done.
> 
> > 1. Is there a particular reason you merged this functionality into FileInputSource rather than making a new TarInputSource? It looks like the two don't actually share much code, and for the methods you didn't add stuff to, it's not clear to me that they're correct.
> 
> TBH because my screen was too small when I was doing this and I didn't see that FileInputSource was a small class I could reimplement ^_^
> 
> > 3. Rather than passing the input path in the --tar-file argument and ignoring the infile argument, I think `--tar-file` (or just `--tar`?) should be a boolean, and we can still use the infile argument for the path.
> 
> I did roughly this but without adding an argument at all - it just detects `.tar` in the filename. Do you think this is OK? It seems easier for bikeshed-web (assuming it preserves the filename, which I'm not sure about) because then hopefully it won't need any updates.

Oh, I just read my email. Peter suggested using tarfile.is_tarfile because PHP may not preserve the file extension. This seems reasonable, so I'll try it out when I have a chance.

kainino0x:
Drafted that.

kainino0x:
@tabatkins ping for re-review! :)

tabatkins:
Sorry for the delay! Will review today. <3

kainino0x:
Thank you! Eager to start trying this out :)

PR #2296, opened 2022-06-13
WPT: Use Stable Channel Browser Logos

bramus:
Polled is the WPT data of the stable releases. Therefore the results overview should also show the logos of those stable channels.


mathiasbynens:
cc @foolip

Issue #2303, opened 2022-06-21
WPT test summary widget double-counts tests named twice


stephenmcgruer:
This could arguably be user (spec writer?) error, but...

I just discovered `<wpt>`, and it's a **really** neat feature that I'm excited to integrate into [Secure Payment Confirmation](https://w3c.github.io/secure-payment-confirmation)'s spec. However, I discovered that the test 'summary' view does not seem to deduplicate test names for its summary numbers, and so effectively restricts one from naming the same test twice in a spec.

<img width="830" alt="image" src="https://user-images.githubusercontent.com/153264/174866290-b06c9682-da8f-4a33-88d9-c61662c43211.png">
(Example from SPC - we only have 48 tests in total, but when I linked the same test file from multiple places it counted each instance separately to get 71 tests instead! And I have no idea what happened to Firefox or Safari...).

A reasonable answer might be 'split up your test files more', but although we are definitely guilty of too much behavior tested in one test, I think it is reasonable for a test file to test more than a single location in a specification.

Perhaps the test summary component could throw the test file names into a `set()` before computing the summary?

stephenmcgruer:
>  Being able to list a test twice (to put a cross-cutting test next to both things it exercises, for example) is absolutely a supported use case, so this is a bug on my part.

Good to hear :). I've sent a PR for this that I hope is correct, so please (when you have time, no rush!) take a look and lmk. I think it does need your permission to run the checks as I am a 'first time contributor' here :D.

PR #2304, opened 2022-06-21
Only count each WPT test once for the summary view

stephenmcgruer:
Previously, every .wpt-name was counted for calculating the pass/total summary,
even if it had previously been seen. This would cause inaccurate results for
specifications where authors list the same WPT test more than once (e.g., if
the test relates to multiple parts of the spec).

Fixes https://github.com/tabatkins/bikeshed/issues/2303

### DIFF
--- a/bikeshed/wpt/wptScript.js
+++ b/bikeshed/wpt/wptScript.js
@@ -23,7 +23,7 @@ document.addEventListener("DOMContentLoaded", async ()=>{
         const passes = result.legacy_status.map(x=>[x.passes, x.total]);
         return [testPath, passes];
     }));
-    const seenTests = Set();
+    const seenTests = new Set();
     document.querySelectorAll(".wpt-name").forEach(nameEl=>{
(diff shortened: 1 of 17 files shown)


bikeshed/wpt/wptScript.js
```diff
@@ -44,6 +41,17 @@ document.addEventListener("DOMContentLoaded", async ()=>{
             })),
         );
         nameEl.insertAdjacentElement("afterend", resultsEl);
+
+	// Only update the summary pass/total count if we haven't seen this
+	// test before, to support authors listing the same test multiple times
+	// in a spec.
+        if (!seenTests.has(nameEl.getAttribute("title"))) {
```
tabatkins, line 48: I know the indentation in this file is inconsistent between tabs and spaces (ugh, I dunno how that happened) but that doesn't mean that mixing 4-space and 2-space indents is allowed as well. ^_^ Mind switching it all to tab-based?

PR #2331, opened 2022-08-10
Import annotations from future to make 'bikeshed update' work on Pythons <3.10.

hober:


tabatkins:
Ah, did this myself this morning, but thanks!

(And btw, this is why everyone should be using the pip install; I *do* rely on my CI to catch mistakes like this, so it's always possible the HEAD is broken at any given point.)

Issue #1664, opened 2020-05-06
Add Remote Playback API to Shepherd


tidoust:
See request to be able to autolink to Remote Playback API terms in https://github.com/w3c/remote-playback/issues/137

The Remote Playback API uses Respec, and was last published to /TR/ in 2017, so current /TR/ document does not respect the definitions contract, and my understanding is that Shepherd cannot easily extract terms from Editor's Drafts that use Respec, so I don't know whether it's possible to add the spec for now.

tidoust:
@tabatkins The Remote Playback API has been re-published since this issue was created (and the Editor's Draft on w3c.github.io is now the generated version of the document). Could the spec now be added to Shepherd for xref purpose, please?

Same question for the Presentation API while I'm at it :)

- Remote Playback API: https://www.w3.org/TR/remote-playback/
- Presentation API: https://www.w3.org/TR/presentation-api/

tidoust:
Gentle ping @tabatkins @plinss. Could the specs be added to Shepherd, please?

plinss:
Added both.

For future reference, Shepherd has a lot of heuristics to determine anchor types even without the definitions contract attributes (it was running before the contract was created).

Also, it's better for Shepherd if the drafts are generated into an Overview.html file rather than index.html to match /TR (it ensures that any links directly to the index page get reduced to a bare path). (It actually doesn't matter what you call, just be consistent between drafts and /TR.)

tidoust:
Thanks @plinss.

The definitions contract does not talk about headings and it seems that Shepherd cannot extract headings from specs generated with ReSpec, e.g.: [Presentation API](https://github.com/tabatkins/bikeshed-data/blob/main/data/headings/headings-presentation-api-1.json), [Payment Request](https://github.com/tabatkins/bikeshed-data/blob/main/data/headings/headings-payment-request-1.json). This makes it impossible to use section links such as [`[[PRESENTATION-API#establishing-a-presentation-connection]]`](https://www.w3.org/TR/presentation-api/#establishing-a-presentation-connection).

What is the markup that Shepherd expects for headings?

plinss:
Shepherd parsed all the headings just fine. I confirmed all the heading archors are in the DB and were properly determined to be section headings. I don't know why they're not listed in the Bikeshed repo, maybe that's just taking time to update. 

[...]

PR #2358, opened 2022-10-05
Cater for ReSpec specs in headings extraction rules

tidoust:
Via https://github.com/tabatkins/bikeshed/issues/1664#issuecomment-1264307993

When a spec generated with ReSpec is added to Shepherd for cross-referencing purpose, definitions properly find their way onto bikeshed-data. Headings do not however (e.g. see [empty headings file of the Presentation API](https://github.com/tabatkins/bikeshed-data/blob/267f66ae75d7ed43852300a63b69ba05ef42bd2e/data/headings/headings-presentation-api-1.json)).

The code that deals with anchor data from Shepherd only considered entries that had both a `heading` type and a `section` flag. Problem is ReSpec wraps sections using `<section>` and adds an ID to both the `<section>` and the heading elements. This makes Shepherd end up with two entries:

- one for the `<section>` of type `other` with a `section` flag
- one for the heading element of type `heading` but without `section` flag

Both entries were skipped. This update makes Bikeshed consider that all entries with a `section` flag are to be added to the list of headings and ignore entries of type `heading` (unless they have a `section` flag too).

This makes Bikeshed select the more straightforward `<section>` ID in specs generated with ReSpec and not the less straightforward heading element ID (e.g. `introduction` instead of `x1-introduction`), which would be selected if Bikeshed only considered anchor entries with a `heading` type instead.

Testing locally, this seems to solve the problem for ReSpec specs referenced in Shepherd and does not seem to have bad consequences on Bikeshed specs, but note I did not run a complete re-generation of the anchor data, so there may exist cases where this update adds additional heading entries that should not be viewed as headings.

### DIFF
--- a/bikeshed/update/updateCrossRefs.py
+++ b/bikeshed/update/updateCrossRefs.py
@@ -134,7 +134,7 @@ def update(path: str, dryRun: bool = False) -> set[str] | None:
                 continue
             if "section" in rawAnchor and rawAnchor["section"] is True:
                 addToHeadings(rawAnchor, specHeadings, spec=spec)
-            if rawAnchor["type"] not in ["heading"]:
+            else:
                 addToAnchors(rawAnchor, anchors, spec=spec)
 
     cleanSpecHeadings(headings)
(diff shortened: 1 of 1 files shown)


tabatkins:
Okay, let me do a complete regen and see if my earlier fears for this approach are still founded.

tabatkins:
Yup, looks good now. A bunch of specs gain headings that previously had none, and a few specs gain a handful of headings that were previously skipped due to the way ReSpec formats things.

PR #2364, opened 2022-10-12
Fix anchors processing

tidoust:
PR #2358 fixed headings but also excluded all non `dfn` terms from processing, resulting in many definitions missing from the anchors data. This update makes sure that all terms are taken into account.

(With my apologies for the hiccup. I completely overlooked `config.dfnTypes` when I removed `"heading"` types, and I was so focused on solving the headings problem that I didn't notice this would affect anchors generation)

Note I'm currently re-generating anchors locally to confirm that this works as intended. Will update once done.

### DIFF
--- a/bikeshed/update/updateCrossRefs.py
+++ b/bikeshed/update/updateCrossRefs.py
@@ -219,7 +219,7 @@ def linearizeAnchorTree(multiTree: list, rawAnchors: list[dict[str, t.Any]] | No
         rawAnchors = []
     # Call with multiTree being a list of trees
     for item in multiTree:
-        if (item["type"] == "dfn") or ("section" in item and item["section"] is True):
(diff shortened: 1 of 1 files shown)


tidoust:
That seems to produce the expected result. At least, with this update, I can build Bikeshed specs that reference IDL terms, events and other similar non `dfn` terms again. And headings remain the ones I would expect to find, regardless of whether the underlying spec was generated with Bikeshed or ReSpec.

tidoust:
Hmm, not fully, some WebIDL definitions (e.g. `boolean`) are missing because the definitions are in headings and appear in Shepherd with the `section` flag set.

tabatkins:
Wow, no, this was entirely my fault, I completely misread my own code when reviewing your earlier patch.

PR #2366, opened 2022-10-13
Bump python version

frivoal:
Bikeshed no longer seems to run fine a python 3.8, so don't advertise that it does.

See for example https://github.com/w3c/w3process/actions/runs/3243873384/jobs/5319223498 This bug goes away by using python 3.10

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -221,7 +221,7 @@ To use bikeshed on [Travis CI](https://travis-ci.org/)'s github integration, you
 ```
 language: python
 python:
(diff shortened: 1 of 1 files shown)


tabatkins:
No, it absolutely does, I just forgot the import that makes type annotations work on this file, and it's not part of my automated testing so it didn't get executed. :/ Fixed now and pushed to pypi.

frivoal:
sorry for the incorrect diagnosis, and thank you for the fix.

Issue #2127, opened 2021-08-21
link-defaults don't canonicalize casing properly


tabatkins:
"dfn" type refs automatically lowercase themselves, so you can refer to them with whatever casing. However, if you set up a link-defaults block referring to a dfn-type ref and use capital letters in its link text, it won't trigger the match.

As an example, cssom-view-1 defined "CSS layout box" without exporting it. If you say:

```
<pre class=link-defaults>
spec: cssom-view-1; type: dfn; text: CSS layout box
</pre>
[=CSS layout box=]
```

...you'll get a link failure. If you lowercase the link-defaults entry to "css layout box" it'll work.

This is broken.

PR #2383, opened 2022-11-03
Add test that canonicalization is correct.

dlaliberte:
The bug reported in #2127 seems to have been fixed. This additional test makes sure that it stays that way.

### DIFF
--- a/tests/links008.bs
+++ b/tests/links008.bs
@@ -1,0 +1,29 @@
+<pre class=metadata>
+Title: Foo
+Group: test
+Shortname: foo
+Level: 1
+Status: LS
+ED: http://example.com/foo
+Abstract: Testing link-defaults canonicalization.
+Editor: Example Editor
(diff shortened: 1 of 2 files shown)


tests/links008.bs
```diff
@@ -0,0 +1,29 @@
+<pre class=metadata>
+Title: Foo
+Group: test
+Shortname: foo
+Level: 1
+Status: LS
```
tabatkins, line 6: FYI, I use the DREAM status for tests to avoid creating all the boilerplate that otherwise makes it very hard to see the important part of the output file. I'll merge this and just change/regen the tests, tho.

tabatkins: ...well, no, I was 100% wrong. This was literally the *only* test using the DREAM status; every single other test in the repo (that isn't specifically testing something about a particular Status) is LS. IGNORE ME.

PR #2387, opened 2022-11-04
Update aiohttp to 3.8.3

jgraham:


jgraham:
No idea if I did the right. `pipenv lock --requirements` doesn't seem to exist now, but `pipenv requirements` does. To get it to update at all I had to delete the lockfile and totally regenerate it.

tabatkins:
Dependabot had a pending update PR for this, so I just merged that since I know things work properly when I do that. ^_^

PR #2406, opened 2022-11-25
Fix algorithm documentation

dlrobertson:
The [algorithm tests](https://github.com/tabatkins/bikeshed/blob/main/tests/algorithm001.bs#L23-L27) do not add a algorithm name and neither do some uses in csswg-drafts. Remove the algorithm tag arguments in the documentation. The current documentation caused some confusion in https://github.com/whatwg/fetch/pull/1549.

Wasn't sure if I should also update the generated documentation html file. Feedback is very much welcomed!

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1651,7 +1651,7 @@ Bikeshed explicitly recognizes this,
 and has several features related to this.
 
 <dfn>Algorithms</dfn> can be explicitly indicated in your markup
-by putting the `algorithm="to foo a bar"` attribute on a container element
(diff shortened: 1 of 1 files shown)


tabatkins:
Ended up writing it a bit differently in 8d83e809b. Thanks for the push!

dlrobertson:
> Ended up writing it a bit differently in [8d83e80](https://github.com/tabatkins/bikeshed/commit/8d83e809b1095a7e65f48f54e5d2c564122f8e1e). Thanks for the push!

Thanks!

Issue #2282, opened 2022-05-20
Add JS for "Pinning" headings/dfns, like JS spec


tabatkins:
In the JS spec, when you hover a heading you get a popup with a few options, one of which is "Pin". Clicking that puts the heading in a special new "Pins" section of the ToC, so you can easily flip back and forth between several relevant sections without having to open multiple windows. This would be cool to add into Bikeshed.

PR #2440, opened 2023-01-03
Implement "Add to Bookmarks" feature

dlaliberte:
This PR implements the "Add to Bookmarks" feature, though it is more of a prototype or proof of concept since I expect many things should be changed.

This addresses [2282](https://github.com/tabatkins/bikeshed/issues/2282)

When hovering over a header, the "Add bookmark" icon is displayed above the header and aligned to the left.  Hovering over that icon shows a tooltip saying "Add bookmark".

image

Clicking the icon adds a bookmarks section to the table of contents and adds a link to the header to an unordered list, which looks like this:

image

Deficiencies:

- The text shown in the bookmark is the id of the header, but should probably be the full text.
- There is no way to remove a bookmark once it has been added.
- There is no memory of bookmarks if the user refreshes the page.

This is implemented as a change only to the bikeshed/boilerplate.py.   Would that be enough?  How do we deal with custom boilerplates?


tabatkins:
Nice!

> The text shown in the bookmark is the id of the header, but should probably be the full text.

Agreed. Just pulling the `.textContent` of the `span.content` element should work.

> There is no way to remove a bookmark once it has been added.

Not an MVP feature but shouldn't be difficult either, just drop a focusable delete icon at the end of the bookmark text.

> There is no memory of bookmarks if the user refreshes the page.

Probably not an issue, since the ECMAScript spec doesn't do this either. If we did add this we'd want to add a little more power, too, like the ability to clear all bookmarks easily. Fine with leaving this out for now.

> This is implemented as a change only to the bikeshed/boilerplate.py. Would that be enough? How do we deal with custom boilerplates?

Yes, this is fine. Boilerplate files are just some HTML; they don't have anything to do with the Python-generated boilerplate stuff here.

I'll leave some inline review comments on the actual code.

bikeshed/boilerplate.py
```diff
@@ -775,18 +775,96 @@ def addIDLSection(doc: t.SpecT) -> None:
         del el.attrib["id"]
     h.addClass(doc, container, "highlight")
 
+bookmarkCss = """
+.material-symbols-outlined {
+  font-variation-settings: 'FILL' 0, 'wght' 400, 'GRAD' 0, 'opsz' 24
```
tabatkins, line 780: Mix of indentations here and in the script. Use a consistent 4-space indent to match the rest of the file, and start the script from zero indent.

bikeshed/boilerplate.py
```diff
@@ -775,18 +775,96 @@ def addIDLSection(doc: t.SpecT) -> None:
...
+        bookmark.setHTML(`<li><a href="#${header.id}">${header.id}</a></li>`);
+        bookmarksList.insertAdjacentElement('beforeend', bookmark);
+    }
+"""
 
 def addTOCSection(doc: t.SpecT) -> None:
     toc = getFillContainer("table-of-contents", doc=doc, default=False)
     if toc is None:
         return
+    # Load icon for "add bookmark"
```
tabatkins, line 851: Hide this and the extraScripts/Style behind an `if "bookmark" not in doc.md.boilerplate` check, so it can be toggled off in the same consistent way other things that add to scripts/styles do.

tabatkins (changes requested):

bikeshed/boilerplate.py
dlaliberte: Doing this is fine, but I haven't yet figured out how to toggle the bookmarks on.

bikeshed/boilerplate.py
```diff
@@ -12,8 +12,8 @@
 from .translate import _
 
 if t.TYPE_CHECKING:
-    MetadataT: t.TypeAlias = t.Mapping[str, t.Sequence[MetadataValueT]]
     MetadataValueT: t.TypeAlias = str | t.NodesT | None
+    MetadataT: t.TypeAlias = t.Mapping[str, t.Sequence[MetadataValueT]]
```
dlaliberte, line 16: I swapped these two lines, since I noticed in the Problems tab in vscode that it said MetadataValueT was undefined.  Not sure this is correct, but at least that problem goes away :)  Let me know what to do otherwise.

bikeshed/boilerplate.py
dlaliberte: Fixed, I think.  The bookmarkCss and bookmarkScript variable are defined at the top level, with content that is now indented.

dlaliberte:
> > There is no way to remove a bookmark once it has been added.
> 
> Not an MVP feature but shouldn't be difficult either, just drop a focusable delete icon at the end of the bookmark text.

Well, I added the ability to remove bookmarks, but I also have to suppress the default click action when clicking the tooltip of the bookmark since it is a link.  Not sure that is working yet.  Since my last changes, I can't view anything on my cloudtop.  

There are also some CI/lint errors I don't understand.

bikeshed/boilerplate.py
tabatkins: Sigh, annoying that VSCode apparently isn't recognizing that the code is intentionally hidden from the Python runtime. The order doesn't matter for mypy and Python doesn't execute that code (since it's under an `if` that's always false at runtime), so there's no need for you to do anything here.

tabatkins:
> There are also some CI/lint errors I don't understand.

Cool, I'll look into these in a bit.

tabatkins:
Ah, it's a `black` complaint - I use `black` to autoformat the code. No worries, I can reformat before merging.

dlaliberte:
> Well, I added the ability to remove bookmarks, but I also have to suppress the default click action when clicking the tooltip of the bookmark since it is a link. Not sure that is working yet. Since my last changes, I can't view anything on my cloudtop.

After two days of trying to get my remote cloudtop working as it was, with no idea why it suddenly stopped working, and two techstop chat sessions that helped narrow down the problem to "Forbidden" access to the remote server, I still don't know why, or how to fix it, so I haven't done any more work on this issue.  Maybe it would be best to just disable the "Remove bookmark" feature for now, which is easy enough.

bikeshed/boilerplate.py
dlaliberte: Still need to toggle bookmarks feature on, but that can be a separate PR.

dlaliberte:
> Ah, it's a `black` complaint - I use `black` to autoformat the code. No worries, I can reformat before merging.

I managed to install the 'black' formatter extension for vscode, and reformatted the document, which resulted in a few changes that now pass the lint tests.

PR #2451, opened 2023-01-29
Fix broken link in docs

sidvishnoi:


sidvishnoi:
As an aside, please consider using spec-prod (or a custom workflow) to auto-generate `docs/index.html` while deploying (instead of asking during PR). Would you like a PR for that?

tabatkins:
> Would you like a PR for that?

Heh, I'd love one.

Issue #2443, opened 2023-01-06
"switch" dl icon and dt permalink icon overlap slightly 


tabatkins:
See <https://w3c.github.io/csswg-drafts/css-position-3/#absolute-cb>, where the pilcrow and the "switch" emoji slightly overlap

PR #2467, opened 2023-02-13
Add padding before switch icon, rebase tests

dlaliberte:
This PR is one way to fix the overlapping icons reported by #2443.

image

All switch dts are indented by this extra padding.
Alternatively, we might outset the pilcrow to the left.


tabatkins:
This seems fine, but you'll have to revert the test rebase, rebase the branch onto main, and re-rebase the tests.

dlaliberte:
> This seems fine, but you'll have to revert the test rebase, rebase the branch onto main, and re-rebase the tests.

I don't know how to do half of that, but I did merge from origin/main, and maybe that did everything I need since rebasing the tests didn't change anything.  I do see the changes re dfnpanels.py in my pilcrow branch, so I think everything should be correct.

PR #2470, opened 2023-02-14
Fix "CanIUse" panel positioning

### LINKED ISSUE #2442: MDN and CanIUse annotations can sit on top of each other
For example at https://drafts.csswg.org/css-cascade-4/#initial, we have:

Screenshot of the "7.3.1. Resetting a Property: the initial keyword" section showing a CanIUse box sitting directly on top of an MDN box

dlaliberte:
This PR should fix #2442.

Looks like this now:

image

I basically made the CanIUse panels be more like the MDN panels. 

1. Move the asides inside the 'main' element.
2. Do repositioning of the asides when window is resized, in addition to after initial loading.
3. Use max-width: min-content, rather than min-width: 1.5em and width: 1em.

There is still one case when the repositioning is missed, but needed: after the showing or hiding of the toc panel.  Is there an event for that, or maybe I should listen for resize on the main window panel, whatever that is.


bikeshed/caniuse/caniuse.py
```diff
@@ -13,6 +13,9 @@ def addCanIUsePanels(doc: t.SpecT) -> None:
     # for a term's feature.
     if not doc.md.includeCanIUsePanels:
         return
+    main = h.findAll("main", doc.body)[0]
```
tabatkins, line 16: This'll throw if there's no main element.

```suggestion
    main = h.find("main", doc.body)
```

PR #2518, opened 2023-04-05
Add some defaults to the template

jyasskin:
And the Repository: line that we now recommend.

This turns off css shorthands, which is debatable, but I figure that they're easy enough to trigger accidentally that most non-CSS specs should have them off, and it's easy enough for CSS specs to turn them back on.

`Assume Explicit For` is also debatable, but without it, I think it's too easy for specs to add new definitions that are correctly scoped, but that break unrelated references.

### DIFF
--- a/bikeshed/cli.py
+++ b/bikeshed/cli.py
@@ -667,7 +667,6 @@ Editor: Your Name, Your Company http://example.com/your-company, your-email@exam
 Abstract: A short description of your spec, one or two sentences.
 Complain About: accidental-2119 yes, missing-example-ids yes
 Markup Shorthands: markdown yes, css no
-Assume Explicit For: yes
 </pre>
 
 Introduction {#intro}
--- a/bikeshed/cli.py
+++ b/bikeshed/cli.py
@@ -661,9 +661,13 @@ Shortname: your-spec
 Level: 1
 Status: w3c/UD
 Group: WGNAMEORWHATEVER
+Repository: org/repo-name
 URL: http://example.com/url-this-spec-will-live-at
 Editor: Your Name, Your Company http://example.com/your-company, your-email@example.com, http://example.com/your-personal-website
 Abstract: A short description of your spec, one or two sentences.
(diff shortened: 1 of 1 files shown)


tabatkins:
I'm against having `Assume Explicit For` by default; when there's an issue Bikeshed will report a link error. `Assume Explicit For` just makes it impossible to hit that link error (at the cost of potentially silently linking to the wrong thing, in the exact case that the link error will fire on).

The rest of the changes look good.

jyasskin:
You're worried about the case when there's both [=/foo=] and [=bar/foo=], and a person writes [=foo=] but meant [=bar/foo=], right?. I'm worried about the case when there's just [=bar/foo=], the person writes [=foo=], and then later someone adds either [=/foo=] or [=baz/foo=]. It's true that the harm from my case is only a new build error, while the harm from your case is a wrongly-directed reference. I think I still disagree, but not enough to argue about it. I've removed the `Assume Explicit For` change.

tabatkins:
Nah it's troublesome either way.

*Without* `Assume Explicit For`, then if an author writes both `<dfn>foo</dfn>` and `<dfn for=bar>foo</dfn>` in their own spec, and then writes `[=foo=]`, they'll get a link error. If they write just `<dfn for=bar>foo</dfn>`, but a *different* spec writes `<dfn>foo</dfn>`, then `[=foo=]` will trigger a *different* link error. In no case will you silently link to the wrong thing; it'll always complain.

*With* `Assume Explicit For`, if the author writes `<dfn for=bar>foo</dfn>` in their spec, but a different spec already has `<dfn>foo</dfn>`, and the author writes `[=foo=]`, it will *silently* link to the external spec (because it assumes you meant `/foo`). Similarly if they define both of them in their own spec, but just write `[=foo=]` without realizing both definitions exist - it'll silently link to the for-less one.

In general `Assume Explicit For` is more dangerous. It exists because Domenic asked for it and it was easy to add. I don't think it should be used in general.

PR #2527, opened 2023-04-17
chore: migrate flake8 to ruff

nschonni:
Noticed this being implemented over on the Node.js project and thought it might make sense here too.
https://beta.ruff.rs/docs/ it doesn't replace all the tools, but I used the recommended migration for the flake8 parts. It's kind of attractive for local development, since the CLI is almost instantaious with this Rust implementation


pyproject.toml
```diff
@@ -1,2 +1,85 @@
...
+    "unsubscriptable-object",            # false positives
+    "unsupported-binary-operation",      # false pos on type sigs, plus mypy catches anyway
+    "use-dict-literal",                  # don't care
+    "useless-return",                    # need it for mypy
+]
+
+[tool.ruff]
+target-version = "py37"
+ignore = [
+    "E501",
```
cclauss, line 52:
```suggestion
```
`E501` and `line-length` both affect the same test.  It is best to set `line-length=447` (below) so that contributors who add even longer lines need to justify their decision in the PR review process.

pyproject.toml
```diff
@@ -1,2 +1,85 @@
...
+    # bare accept is fine
```
cclauss, line 53:
```suggestion
    # bare except is fine
```

cclauss (approved):
Nice!

pyproject.toml
```diff
@@ -1,2 +1,85 @@
...
+    "N803",
+    "N806",
+    "N812",
+    "N815",
+    "N816",
+    "N999",
+]
+line-length = 120
+select = [
+    "E",
```
cclauss, line 77: Add the functionality of `rbialon/flake8-annotations`...
```suggestion
    "ANN",
    "E",
```

Also, `"PL"` for pylint

pyproject.toml
tabatkins: Nah, I still think max line lengths are a fake idea. `black` enforces reasonable line-breaking practices by itself; the remaining long lines are due to long strings, and I don't think that manually breaking those into separate strings is worth anyone's time or effort. (In fact I think it's *actively a bad idea*, as changes to the string might require cascading edits to rejigger the line lengths. I follow [Semantic Line Breaks](https://sembr.org/) in most of my projects, and where that doesn't make sense, rely on soft-wrapping.

PR #2532, opened 2023-04-20
Initial install of playwright

dlaliberte:
The default install of playwright mostly works.  I chose 'JavaScript' and tests are in tests/e2e.  The webkit tests are disabled because I get missing dependency errors. Several simple tests are included which generate screenshots.


.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ main, master ]
+  pull_request:
+    branches: [ main, master ]
+jobs:
+  test:
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
```
nschonni, line 10: I'm guessing the webkit stuff isn't working because it would only run if you add a `macos-latest` run to the matrix where Safari would be available on the runner

dlaliberte: That sounds reasonable.  I'll let you or others experiment after I land this PR.

PR #2533, opened 2023-04-21
chore: remove guards and EOL requirements.txt deps

nschonni:
- Remove guards for >= 3.x that were all or equal to the minimum version
- removed six and urllib3 that are no longer imported

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,33 +6,31 @@
 #
 
 -i https://pypi.org/simple
-about-time==4.2.1; python_version >= '3.7' and python_version < '4'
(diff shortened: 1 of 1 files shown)


tabatkins:
This file is now auto-genned; those guards are added by pipenv. If you can figure out how to make *it* not generate those, I'd be fine with that, but just editting *this* file will get overwritten next time I update requirements.

(But I also don't think it's important, so don't spend time on it if you don't immediately know how to do this with pipenv.)

I'll get six and urllib3 removed properly, thanks for discovering that.

PR #2601, opened 2023-07-10
Run playwright ci action in docker

dlaliberte:
This PR adds a separate playwright.yml to run playwright in a docker image.  It also adds node commands to run playwright in the same docker image on the developer's machine.


package.json
```diff
@@ -8,8 +8,8 @@
     "test": "tests"
   },
   "scripts": {
-    "e2e": "playwright test",
-    "e2e:report": "playwright show-report",
+    "pwtests": "docker run --rm --network host -v $(pwd):/work/ -w /work/ -it mcr.microsoft.com/playwright:v1.35.1-jammy /bin/bash -c \"npm ci; npx playwright install; npx playwright test\"",
```
jyasskin, line 11: Comment that the version of the image here has to exactly match the versions of @playwright/test in devDependencies and the Github Actions configuration. The version in devDependencies should just be "1.35.1" without the `^`, so it doesn't auto-upgrade and get out of sync. You should probably remove `playwright-webkit` from "dependencies" entirely.

You also should avoid running `npx playwright install`: the image includes that data, and we want this to fail quickly if the versions get out of sync, instead of doing the slow download of all the browser versions.

The commands should probably be separated with `&&` instead of `;` so that later ones don't run if earlier ones fail.

Finally, make sure that all of those docker flags are actually needed, and probably spell out their --long-forms instead of the single-character option names. https://docs.docker.com/storage/bind-mounts/ argues to replace -v with --mount.

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,65 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ main, master ]
+  pull_request:
+    branches: [ main, master ]
```
jyasskin, line 6: There's no master branch, and this should probably run in the same cases as the main CI workflow:
```suggestion
  push:
  push:
    branches-ignore:
      - "dependabot/**"
      - "daily-test-sync"
  pull_request:
```

Arguably, this should actually just be a second job inside that main `ci.yml`...

jyasskin (changes requested):

package.json
dlaliberte: Comments in json are not allowed, so I added a pwtests-comment command :).  
Need to look up -it, --rm, -w, and -c to find their long-form names.

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,35 @@
...
+        - uses: actions/checkout@v3
+
+        # Steps for Playwright testing with container
+        - uses: actions/setup-node@v3
+          with:
+            node-version: 18
+        - name: Install dependencies
+          run: |
+            npm ci
+            npx playwright install
```
jyasskin, line 27: Ah, this also shouldn't install the browsers since they're supposed to be provided by the docker image:
```suggestion
          run: npm ci
```

package.json
```diff
@@ -8,8 +8,9 @@
     "test": "tests"
   },
   "scripts": {
-    "e2e": "playwright test",
-    "e2e:report": "playwright show-report",
+    "//pwtests": "The version of the docker playwright image used here must exactly match the versions of @playwright/test in devDependencies and the Github Actions configuration (in playwright.yml).  The version of playwright used in devDependencies should not include the ^ prefix, so it doesn't auto-upgrade and get out of sync.",
```
jyasskin, line 11: Oh, I'd forgotten `package.json` was a format that doesn't have comments. No strong opinion between keeping this and removing it.

jyasskin (approved):

package.json
dlaliberte: I think the comment about consistent versions is very important.  We could stuff it in playwright.yml instead, but it is more likely to be found here.

PR #2615, opened 2023-07-21
Dockerfile: Copy semver.txt to /app/bikeshed, not /app

rakuco:
This follows the changes that landed a few months ago and moved
semver.txt to the bikeshed/ subdirectory.

While here, remove the top-level semver.txt that does not seem to be
used anymore.


rakuco:
cc @tabatkins

PR #2617, opened 2023-07-24
update: Pass trust_env=True when creating aiohttp.ClientSession instances

rakuco:
aiohttp does not read environment variables (such as `$http_proxy`) by
default, which causes `bikeshed update` to time out when run behind a
proxy.

Explicitly pass trust_env=True to make it use proxy environment
variables in HTTP request.


rakuco:
cc @tabatkins

tabatkins:
Ooh, good catch.

PR #2628, opened 2023-08-01
Let the patcg-id "group" use the UD status.

jyasskin:
Per https://github.com/speced/bikeshed-boilerplate/pull/43#issuecomment-1658832017, it should be possible for documents with `Group: patcg-id` to use the `UD` status, which isn't allowed for groups that are in `w3cCgs`. I think it's right to keep this "group" in `megaGroups["w3c"]`, although I'm not certain.


tabatkins:
k, fixed and released in 3.14.4

Issue #1319, opened 2018-07-25
Show how to link to a term


jyasskin:
For example, when trying to link to https://mimesniff.spec.whatwg.org/#mime-type-essence, I wind up using Chrome's inspector to reveal that the element is 

```html
<dfn class="dfn-paneled" data-dfn-for="MIME type" data-dfn-type="dfn"
data-export="" id="mime-type-essence">essence</dfn>
```

From which I conclude that the right link is `[=MIME type/essence=]` and that there aren't any other `lt`s I can use to refer to it.

It'd be nice to be able to see that information from the dfn panel. However, nobody but other spec authors cares, so it probably shouldn't be shown by default.

tabatkins:
Right, that's about the best way to figure out term linking right now. Alternately one can use use `bikeshed refs`:

```
> bikeshed refs --text="essence"
status:    current
for:       [MIME type]
level:     1
url:       https://mimesniff.spec.whatwg.org/#mime-type-essence
type:      dfn
export:    True
text:      essence
shortname: mimesniff
normative: True
spec:      mimesniff
```

But yeah, exposing this data somehow would be interesting. @jakearchibald is asking for a vaguely-related "power feature" for definitions in #1321; maybe I could put some icons at the bottom of the dfn panels?

jyasskin:
Unfortunately, `bikeshed refs --text contains` doesn't discover that https://fetch.spec.whatwg.org/#header-list-contains is also accessible via `[=header list/does not contain=]`.

Icons with tooltips sound fine for power features.

FWIW, #1321 sounds like a good behavior for the *default* click on a term: a lot of times non-power-users don't need to lose their context in the source standard just to figure out what a term meant.

PR #2685, opened 2023-10-03
Dfn panel changes for linking syntax

dlaliberte:
This PR adds a "Possible linking syntaxe(es)" section to dfn panels.  

* The copy icon is perhaps overkill.
* The cursor for the dfn is changed to be a 'help' cursor, since the dfn is not a link to anything but the dfn panel.
* There are also a couple css changes regarding dfn panels in general, to maybe fix some whitespace issues, depending on what was intended.
  * Avoid wrapping after section numbers. 
  * No ellipsis needed most of the time.
  * Multiple reference numbers no longer wrap (e.g. (2) (3) (4))

Here is what it looks like now.  

image


bikeshed/dfnpanels/dfnpanels.js
```diff
@@ -226,7 +312,6 @@
                 pinDfnPanel(dfnPanel);
             }
             event.stopPropagation();
-            event.preventDefault();
```
dlaliberte: Without event.preventDefault(), the page scrolls so that the dfn is at the top of the window, and the dfn panel *appears* to just disappear, but it is moved simultaneously to the bottom left corner.   If that is your intention, fine.  I found it confusing, however.

dlaliberte: I hadn't realized that this event.preventDefault() call was also disabling all the other links on the panel.  We'll look into fixing the scrolling behavior perhaps a different way.

dlaliberte:
This PR fixes #1319

Issue #2645, opened 2023-08-21
Implementation Report metadata not documented


tabatkins:
"Implementation Report" is required for some statuses; it's intended to take a URL to the spec's implementation report (a manually-written document, or just a link to the spec's folder on wpt.fyi). However, this isn't documented anywhere; Bikeshed just complains about its lack with no further information.

PR #2765, opened 2024-01-13
Document (sometimes) required metadata for implementation report and CR deadline

### LINKED ISSUE #2644: CR requires Deadline now?
Currently only FPWD and PR are marked as "needing Deadline" in Bikeshed. However, it appears that CRs now require it as well; Specberus appears to demand it. Also, the CSSWG status.include file has deadline text, but for some reason the macro is marked as optional. (And the line predates me splitting the boilerplates out to their own project, so archaeology will be annoying.)

I need to investigate what the actual requirements are here and get things synced up.

svgeesus:
This PR would fix #2644 and #2645. Links to pubrules in cases where this is required.

Contributing requires the html to be generated. This has a bunch of fatal errors, which I didn't introduce, so I used -f.

```
chris@SuperNomad:/mnt/c/Users/chris/Documents/GitHub/mybikeshed/docs$ bikeshed -f spec
FATAL ERROR: Found unmatched text macro [LOGO]. Correct the macro, or escape it with a leading backslash.
FATAL ERROR: Found unmatched text macro [WORKSTATUS]. Correct the macro, or escape it with a leading backslash.
FATAL ERROR: Found unmatched text macro [FOO]. Correct the macro, or escape it with a leading backslash.
FATAL ERROR: Found unmatched text macro [BAZ]. Correct the macro, or escape it with a leading backslash.
FATAL ERROR: Found unmatched text macro [H1]. Correct the macro, or escape it with a leading backslash.
FATAL ERROR: Found unmatched text macro [LATEST]. Correct the macro, or escape it with a leading backslash.
FATAL ERROR: Found unmatched text macro [ABSTRACTATTR]. Correct the macro, or escape it with a leading backslash.
FATAL ERROR: Found unmatched text macro [DEADLINE]. Correct the macro, or escape it with a leading backslash.
LINE 4114: Dfn has no linking text:
<dfn bs-line-number="4114"><code bs-line-number="4114" bs-autolink-syntax="``" bs-opaque=""></code></dfn>
LINE 4117: Dfn has no linking text:
<dfn bs-line-number="4117"><code bs-line-number="4117" bs-autolink-syntax="`&lt;p bs-line-number=-1&gt;Bikeshed is a spec-generating tool that takes in lightly-decorated Markdown and spits out a full spec, with cross-spec autolinking, automatic generation of indexes/ToC/etc, and many other features.&lt;/p&gt;
`" bs-opaque=""></code></dfn>
LINE 4120: Multiple elements have the same ID 'macro-13-january-2024'.
Deduping, but this ID may not be stable across revisions.
LINE 4120: Multiple local 'dfn' <dfn>s for 'macro' have the same linking text '13 january 2024'.
LINE ~4124: No 'dfn' refs found for '[date]'.
[=[DATE]=]
 ✔  Successfully generated, but fatal errors were suppressed
```

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -1488,9 +1488,9 @@ Possible extra rowspan handling
 		}
 	}
 </style>
-  <meta content="Bikeshed version 815accb43, updated Thu Dec 21 14:10:37 2023 -0800" name="generator">
+  <meta content="Bikeshed version 82ce88815, updated Thu Sep 7 16:33:55 2023 -0700" name="generator">
   <link href="https://speced.github.io/bikeshed/" rel="canonical">
-  <meta content="815accb43c1370ab476546acca5e80f08024a1f8" name="revision">
+  <meta content="178fa41ecc3c94893c1692be95ea728e0bcf049b" name="document-revision">
 <style>/* Boilerplate: style-autolinks */
 .css.css, .property.property, .descriptor.descriptor {
     color: var(--a-normal-text);
@@ -1562,6 +1562,7 @@ pre .property::before, pre .property::after {
 }
 </style>
 <style>/* Boilerplate: style-colors */
+
 /* Any --*-text not paired with a --*-bg is assumed to have a transparent bg */
 :root {
     color-scheme: light dark;
@@ -1815,15 +1816,11 @@ figcaption:not(.no-marker)::before {
 :root {
     --dfnpanel-bg: #ddd;
     --dfnpanel-text: var(--text);
-    --dfnpanel-target-bg: #ffc;
-    --dfnpanel-target-outline: orange;
 }
 @media (prefers-color-scheme: dark) {
     :root {
         --dfnpanel-bg: #222;
         --dfnpanel-text: var(--text);
-        --dfnpanel-target-bg: #333;
-        --dfnpanel-target-outline: silver;
     }
 }
 .dfn-panel {
@@ -1846,13 +1843,11 @@ figcaption:not(.no-marker)::before {
 .dfn-panel > b { display: block; }
 .dfn-panel a { color: var(--dfnpanel-text); }
 .dfn-panel a:not(:hover) { text-decoration: none !important; border-bottom: none !important; }
-.dfn-panel a:focus {
-    outline: 5px auto Highlight;
-    outline: 5px auto -webkit-focus-ring-color;
-}
 .dfn-panel > b + b { margin-top: 0.25em; }
 .dfn-panel ul { padding: 0 0 0 1em; list-style: none; }
 .dfn-panel li a {
+    white-space: pre;
+    display: inline-block;
     max-width: calc(300px - 1.5em - 1em);
     overflow: hidden;
     text-overflow: ellipsis;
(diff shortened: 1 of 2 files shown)


svgeesus:
@tabatkins could you review this PR please? Closes a couple of issues.

docs/index.html
```diff
@@ -6045,58 +6049,61 @@ <h3 class="heading settled" data-level="12.3" id="text-macros"><span class="secn
...
     <li data-md>
-     <p><dfn class="dfn-paneled" data-dfn-for="macro" data-dfn-type="dfn" data-export id="macro-h1"><code>[H1]</code></dfn> gives the desired document heading, in case the in-page title is supposed to be different from the <code>&lt;title></code> element value.</p>
+     <p><dfn data-dfn-for="macro" data-dfn-type="dfn" data-export id="macro-h1"><code>[H1]</code><a class="self-link" href="#macro-h1"></a></dfn> gives the desired document heading, in case the in-page title is supposed to be different from the <code>&lt;title></code> element value.</p>
     <li data-md>
-     <p><dfn class="dfn-paneled" data-dfn-for="macro" data-dfn-type="dfn" data-export id="macro-shortname"><code>[SHORTNAME]</code></dfn> gives the document’s shortname, like "css-cascade".</p>
+     <p><dfn data-dfn-for="macro" data-dfn-type="dfn" data-export id="macro-bikeshed"><code>bikeshed</code><a class="self-link" href="#macro-bikeshed"></a></dfn> gives the document’s shortname, like "css-cascade".</p>
```
ferdnyc, line 6056: The HTML-generation process seems to have gone very wrong here — the newly-generated HTML is actually replacing the macros, instead of documenting them. It seems strings surrounded by `[` and `]` are getting replaced even when inside `<code>`, where previously they were left alone.

docs/index.html
```diff
@@ -4429,7 +4433,7 @@ <h3 class="heading settled" data-level="5.11" id="including"><span class="secno"
 </pre>
    <p>With the above code, you can use <code>[FOO]</code> and <code>[BAZ]</code> macros inside the include file,
 and they’ll be substituted with "bar" and "qux qux qux", respectively.
-(Remember that you can mark text macros as optional by appending a <code>?</code>, like <code>[FOO?]</code>,
+(Remember that you can mark text macros as optional by appending a <code>?</code>, like <code></code>,
```
ferdnyc, line 4436: Same problem here, with (ostensible) macros now being replaced inside literals — though, interestingly, `[FOO]` and `[BAZ]` above still weren't replaced.

docs/index.html
```diff
@@ -3688,30 +3692,30 @@ <h2 class="heading settled" data-level="4" id="metadata"><span class="secno">4.
...
+     <p><dfn class="dfn-paneled" data-dfn-for="metadata" data-dfn-type="dfn" data-export id="metadata-local-boilerplate">Local Boilerplate</dfn> tells Bikeshed which boilerplate include files to look for relative to the specification (as opposed to inside the Bikeshed source). Its value is a comma-separated list of include file basename and <a data-link-type="dfn" href="#boolish" id="ref-for-boolish⑨">boolish</a> value, e.g. <code>Local Boilerplate: footer yes</code> to look for <code>footer.include</code> and <code>footer-LS.include</code> next to the specification. This also works for default includes that aren’t quite boilerplate like <a href="#default-metadata">defaults.include</a>, <a href="#computed-metadata">computed-metadata.include</a>, and <a href="#echidna-hooks">bs-extensions.include</a>.</p>
```
ferdnyc, line 3714: Another undesirable macro-inside-code-element replacement.

docs/index.html
```diff
@@ -3655,13 +3659,13 @@ <h2 class="heading settled" data-level="4" id="metadata"><span class="secno">4.
...
     <li data-md>
      <p><dfn class="dfn-paneled" data-dfn-for="metadata" data-dfn-type="dfn" data-export id="metadata-repository">Repository</dfn> indicates the repository the spec is being tracked in.
 You can specify a url followed by a "short name" for it,
-and it populates the <code>[REPOSITORY]</code> and <code>[REPOSITORYURL]</code> text macros accordingly.
+and it populates the <code>speced/bikeshed</code> and <code>https://github.com/speced/bikeshed</code> text macros accordingly.
```
ferdnyc, line 3668: Here again.

docs/index.html
ferdnyc: Even more confounding, there's **nothing** different about those two uses. The source for this paragraph, in `index.bs`, is:

```markdown
With the above code, you can use `[FOO]` and `[BAZ]` macros inside the include file,
and they'll be substituted with "bar" and "qux qux qux", respectively.
(Remember that you can mark text macros as optional by appending a `?`, like `[FOO?]`,
in which case they'll be replaced with the empty string if Bikeshed can't find a definition.)
```

ferdnyc (commented):
Unfortunately, the HTML generation seems to have drastically affected how macros are documented, since many of them ended up getting replaced when they shouldn't be (or, aren't intended to be).

docs/index.html
ferdnyc: I guess the difference is that `FOO` and `BAZ` are undefined macros, whereas all of the ones that were replaced were defined. (Even if, as in the case of `FOO?`, they were defined to be the empty string.)

ferdnyc:
I wonder if #2721 is responsible for the change in behavior here?

ferdnyc:
I have a fix for the macro-reference issues in #2773 that will allow the docs to be regenerated correctly, and without `-f`. Once that's merged you should be able to resolve any conflicts that arise in `docs/index.bs`, then run `bikeshed` in that directory to cleanly update the HTML version.

tabatkins:
Uh, none of these problems are reproduced in the current version of Bikeshed. I have to assume that Chris's version is substantially behind?

tabatkins:
Okay, committed the index.bs changes (and regened index.html myself). Thanks, @svgeesus !

PR #2773, opened 2024-01-25
Fix macro references in Bikeshed documentation

ferdnyc:
Since 16c53d7 the syntax <code>\`[MACRO]\`</code> is no longer protected from replacement, and an escaped <code>\`\\[MACRO]\`</code> reference will be output as `<code>\[MACRO]</code>`.
So, replace all <code>\`[MACRO]\`</code> references with `<code>\[MACRO]</code>` in the bikeshed documentation source, which is correctly translated into the output document as `<code>[MACRO]</code>`.

The regenerated `docs/index.html` (included) differs only by code changes made to bikeshed's HTML generation boilerplate since the last time it was generated.

Affects #2765 (specifically, the `docs/index.html` generated there)

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1187,7 +1187,7 @@ There are several additional optional keys:
 	* <dfn>Canonical URL</dfn> defines the canonical URL (as used by search engines) for the document; it can be set to "TR" (the default value if TR is defined), "ED", or a specific URL.
 	* <dfn>Favicon</dfn> defines the favicon URL for the document; it only supports one single URL.
 	* <dfn>Logo</dfn> defines the logo url for the document; it only supports one single url.
-		This URL can be used via the `[LOGO]` text macro ([[#text-macros]]),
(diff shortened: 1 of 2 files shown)


ferdnyc:
Ironically, the only reference that doesn't have to be adjusted is this one:

https://github.com/speced/bikeshed/blob/c20becf03fd776947976073d499f478a613359eb/docs/index.bs#L4138

As the new behavior is the same as the old behavior in that particular instance.

tabatkins:
This problem doesn't reproduce in the current version of Bikeshed. I have to assume the issue is the docs being generated with an old version.

ferdnyc:
@tabatkins 

> This problem doesn't reproduce in the current version of Bikeshed. I have to assume the issue is the docs being generated with an old version.

Agreed, though I don't know how as I was working in a venv with a `pip install -e .` version of `bikeshed`.

\*shrug\* Oh, well. Apologies for the noise.

Issue #2705, opened 2023-11-02
Can't describe `required [EnforceRange] typename attributeName`


padenot:
This stems from https://github.com/w3c/webcodecs/pull/488

## STR
- Have a fragment like this:

```
<xmp class='idl'>
dictionary PlaneLayout {
  required [EnforceRange] unsigned long offset;
  required [EnforceRange] unsigned long stride;
};
</xmp>

: <dfn dict-member for=PlaneLayout>offset</dfn>
:: The offset in bytes where the given plane begins within a {{BufferSource}}.

: <dfn dict-member for=PlaneLayout>stride</dfn>
:: The number of bytes, including padding, used by each row of the plane within
    a {{BufferSource}}.
```

## Expected
- it works, everything is linked appropriately

## Actual

```
$ bikeshed --die-on=warning spec index.src.html index.html
LINK ERROR: No 'idl-name' refs found for '[EnforceRange] unsigned long'.
<a data-link-type="idl-name" data-lt="[EnforceRange] unsigned long">[EnforceRange] unsigned long</a>
 ✘  Did not generate, due to errors exceeding the allowed error level.
make: *** [index.html] Error 2
```

`bikeshed` attempts to reference a type that's `[EnforceRange] unsigned long` and that's not a thing.

Instead, we have to write:

```
<xmp class='idl'>
dictionary PlaneLayout {
  [EnforceRange] required unsigned long offset;
  [EnforceRange] required unsigned long stride;
};
</xmp>

: <dfn dict-member for=PlaneLayout>offset</dfn>
:: The offset in bytes where the given plane begins within a {{BufferSource}}.

: <dfn dict-member for=PlaneLayout>stride</dfn>
:: The number of bytes, including padding, used by each row of the plane within
    a {{BufferSource}}.
```

with `[EnforceRange]` and `required` swapped, which is incorrect WebIDL.

PR #2790, opened 2024-02-09
Drop IDL extended attributes from data-type values

tidoust:
This is a fix for #2705, where the problem is that Bikeshed dies on argumentdef tables when a spec defines an IDL interface attribute or a dictionary member that comes with extended attributes, something like

```
dictionary test {
	required [EnforceRange] short member1;
}
```

To produce the ", of type foo" content that appears next to the term in the table, Bikeshed relies on the `data-type` attribute set in the IDL block. The value of that attribute comes from a `type` field returned by widlparser. The code in Bikeshed expected that value to be a real IDL type (a simple type or a union) but that construct in widlparser is a "type with extended attributes": https://github.com/plinss/widlparser/blob/6c91beb68806fd6c206c57cecd15e455b0b6c1c1/widlparser/productions.py#L1294

As a result, the type Bikeshed sees is `[EnforceRange] short` in the above example, which cannot be directly associated with a type in Web IDL.

This update gets back to actual IDL types in the `data-type` attribute, which effectively solves the problem of the ", of type foo" content generation. A new idl008 test was added to check extended attributes.

One side effect of that update is that `data-type` attribute values no longer contain trailing spaces. That seems a good thing (spaces where already stripped for attributes, but not for dictionary members), but required re-generating a bunch of reference tests.

Note: If the inclusion of extended attributes in the `data-type` attribute was actually intended, then the fix would rather need to be done in `attributeInfo.py`.


tabatkins:
Hrm, I wish there was an easier way to just obtain the type without the attributes, but the design of widlparser means that I can't just construct a `Type` without giving it tokens to build itself out of. I'd have to either do a hacky mock tokenizer, or significantly refactor to make all the `__init__` methods just simple constructors, and swap the existing parser-based ones to static functions.

Thanks for the patch!

PR #2801, opened 2024-03-01
Adding Alliance for Open Media specification statuses

podborski:
The [Alliance for Open Media](https://aomedia.org) is using bikeshed in several specifications and these are the statuses we defined.


podborski:
@cconcolato FYI

Issue #2157, opened 2021-10-28
Adapt the var-highlight colors for dark mode


tabatkins:
They're all pastel right now, which looks great on a white background with black text, but is unreadable for dark background with white text.

kainino0x:
@greggman suggests this simple fix might work:

> ```css
> @media (prefers-color-scheme: dark) {
>   var[data-var-color] {
>      color: black;
>      filter: invert(1);
>   }
> }
> ```

PR #2848, opened 2024-05-06
Pick some dark `<var>` colors

greggman:
It seems okay to use oklab now? The comment said the original
values had a and b go from [-128 to +128] so scaled them
to oklab which goes from [-100% to 100%]. Colors 2 and 3 seemed
too close to me so I adjusted color 3 slightly.

[Example](https://jsgist.org/?src=c0db656fdeb5dabe826d2a14860e3c18)

Screenshot 2024-05-06 at 15 07 38

#2157


greggman:
@tabatkins ping?

tabatkins:
And released 4.1.10 with the new styles.

PR #2862, opened 2024-05-23
Add the htmlwg as a w3c group

frivoal:
There's already html, but that corresponds to the old https://www.w3.org/groups/wg/html, which is distinct from https://www.w3.org/groups/wg/htmlwg/


frivoal:
@tabatkins Can you cut a new release of bikeshed to get this out?

PR #2882, opened 2024-06-27
use absolute import in __main__ to support direct use of pyinstaller

jcbhmr:
goal: allow users to create an exe/bundle/whatever bikeshed pip package directly using PyInstaller without undue difficulty.

problem: since the current `bikeshed/__main__.py` uses relative imports it's not directly usable with pyinstaller. you instead need to create your own wrapper package around bikeshed that calls into the bikeshed.cli.main() function yourself:

```sh
pip install bikeshed
```

```py
# mybikeshedhack/__main__.py
from bikeshed.cli import main

main()
```

```sh
pyinstaller -y --collect-all=bikeshed --name=bikeshed mybikeshedhack/__main__.py
```

why is this wrapper so painful? it's not. this is in reality a very tiny problem but since it's also a very tiny change I figured id open a pr. 🤷‍♀️

but it does 1) creating a venv 2) installing bikeshed as a dep (and managing it in like pyproject.toml or requirements.txt or whatever) and 3) remembering the incantation to start bikeshed.cli.main()

solution: use absolute imports in the `bikeshed/__main__.py` file so that you can do this without that extra wrapper.

```sh
pip install bikeshed # or pipx if you can find the venv
# then just use the raw __main__.py from bikeshed directly! no wrapper file!

pyinstaller -y --collect-all=bikeshed ...wherever_venv.../bikeshed/__main__.py

# OR dynamically
pyinstaller -y --collect-all=bikeshed "$(python -c "from importlib.util import find_spec; print(find_spec('bikeshed.__main__').origin)")"
```

### DIFF
--- a/bikeshed/__main__.py
+++ b/bikeshed/__main__.py
@@ -1,3 +1,3 @@
-from .cli import main
+from bikeshed.cli import main
 
 main()


tabatkins:
Sorry for the delay, I'd been on vacation for a while.

I'm not necessarily opposed to this, I'm just not sure what it's doing. All the other files use package-relative imports as well; why is it only the `__main__.py` that needs changing?

Is this a common problem for pyinstaller packages, and this fix a common way to correct it? Any docs or discussion about it?

jcbhmr:
The reason PyInstaller doesn't like relative imports in its input files is a bit murky to me too, but I think it has to do with it just pulling it out and ignoring the context of where said input file was found. Almost like it's in a pseudo `/tmp/inputfile123.py` instead of `/workspace/bikeshed/src/__main__.py`.

There's [lots of Google results](https://www.google.com/search?q=pyinstaller+relative+imports) for "PyInstaller can't find relative import in main.py" and a lot of them (at least the few that I read to solve my problem of creating a PyInstaller .exe for bikeshed) seem to say the same thing: PyInstaller needs absolute imports **for the input file only**. To make this work with the Pythonic `__main__.py` that means using absolute instead of relative imports in `__main__.py`. 🤷‍♂️

PyInstaller issue: https://github.com/pyinstaller/pyinstaller/issues/2560
Solution direct link comment: https://github.com/pyinstaller/pyinstaller/issues/2560#issuecomment-377917879

> Note that's it's not necessary to use absolute imports everywhere. It's sufficient to use absolute imports in the script you're supplying to PyInstaller, only.

---

This could also be done by doing `pyinstaller ./bikeshed.py` at the root of this repository but `./bikeshed.py` isn't included in the PyPI `pip install bikeshed` distribution.

jcbhmr:
An update on the use case: I have solved my problem which was getting a PyInstaller build of Bikeshed for use in places without pip & Python: I fork & build _from the source code_ instead of _from the pip install_ which means I can use the bikeshed.py at the root of the repo instead of the __main__.py from the pip package. I then build binaries in GitHub Actions and publish them as releases. I even combined four PyInstaller builds into a single self-extracting actually portable executable. https://github.com/jcbhmr/bikeshed

PR #2937, opened 2024-10-18
update /standards/types to /standards/types/ per linkchecker warning

tantek:
Linkchecker warns that links to https://www.w3.org/standards/types are redirected to https://www.w3.org/standards/types/ so fix the generation of these links to automatically avoid this warning.

### DIFF
--- a/bikeshed/metadata.py
+++ b/bikeshed/metadata.py
@@ -356,25 +356,25 @@ class MetadataManager:
             statusName = doc.doctype.status.name
             if statusName == "FPWD":
                 macros["w3c-stylesheet-url"] = "https://www.w3.org/StyleSheets/TR/2021/W3C-WD"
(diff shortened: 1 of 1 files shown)


Ms2ger:
I'm getting complaints seemingly in the opposite direction: https://labs.w3.org/echidna/api/status?id=03aca09e-4476-4b3d-933a-4cb62dd006c1

tantek:
@Ms2ger looks like a specberus bug. Could you file it with that error message and this PR since Linkchecker was actually checking the link whereas it looks like specberus has an out of date rule that needs updating?

Ms2ger:
Thanks for the pointer, seems like it's https://github.com/w3c/specberus/issues/1887

anssiko:
(Cross-linking: the Specberus bug https://github.com/w3c/specberus/issues/1887 was fixed by https://github.com/w3c/specberus/pull/1889)

PR #2981, opened 2024-12-06
Update references to "IEEE 754" to the latest version.

sunfishcode:
Add a reference for the current version of IEEE 754, which is "IEEE 754-2019", from the previous version which was "IEEE 754-2008".


tabatkins:
This data comes from the SpecRef project, so I've gone ahead and made the update over there: <https://github.com/tobie/specref/pull/838>.

Closing this PR, since it'll get picked up automatically by `bikeshed-data` in a little bit.

PR #3032, opened 2025-02-12
Fix Candidate Recommendation Draft metadata checks

### LINKED ISSUE #3031: All Bikeshed CRD builds fail without Deadline metadata set
The latest Bikeshed started failing when building a CRD without `Deadline` metadata. Reproduced with multiple specs. One example [via spec-prod run](https://github.com/w3c/accelerometer/actions/runs/13281153288/job/37079618856) omitting warnings for brevity:

```
$ bikeshed --die-on=fatal spec "index.bs" "index.bs.built.html" --md-status="CRD" --md-shortname="accelerometer" --md-date="2025-02-12" --md-Prepare-For-TR="yes" --md-previous-version="https://www.w3.org/TR/2024/CRD-accelerometer-20241009/"
      FATAL ERROR: Not all required metadata was provided:
          Missing 'Deadline'
✘  Did not generate, due to errors exceeding the allowed error level.
```

The process does not require Deadline for CRD (only for CRS), so there's a bunch of specs that don't define `Deadline` metadata for CRDs and it looks like all those builds are now failing.

This seem to be a regression: here's a successful build using the same bs source and Bikeshed 4.1.8 I happened to have around (omitting linking errors):

```
$ bikeshed --die-on=fatal spec "index.bs" "index.bs.built.html" --md-status="CRD" --md-shortname="accelerometer" --md-date="2025-02-12" --md-Prepare-For-TR="yes" --md-previous-version="https://www.w3.org/TR/2024/CRD-accelerometer-20241009/"
 ✔  Successfully generated, with 83 linking errors
```

As a workaround, we could add made-up `Deadline` metadata to all the affected specs, but before going there I'd like to understand if this is a Bikeshed regression that could be fixed, or a bug elsewhere in the spec-prod toolchain?

cc @tabatkins for Bikeshed-fu!

@deniak I discovered this with your PR https://github.com/w3c/accelerometer/pull/83 that bumped the spec-prod runner to ubuntu-latest. Could this be some incompat with the Python version in ubuntu-latest and Bikeshed?  I noticed a bunch of similar PRs for other repos with CRDs, so I won't merge those for the time being to not break too many builds.

Thanks!

anssiko:
CRD does not require this deadline information per process: https://www.w3.org/policies/process/#publishing-crud

Without this fix CRD builds without "Deadline" metadata will fail.

Fix #3031

### DIFF
--- a/bikeshed/spec-data/readonly/boilerplate/doctypes.kdl
+++ b/bikeshed/spec-data/readonly/boilerplate/doctypes.kdl
@@ -188,7 +188,7 @@ org "w3c" {
         group-types "wg"
     }
     status "CRD" "W3C Candidate Recommendation Draft" {
-        requires "Level" "ED" "TR" "Issue Tracking" "Date" "Implementation Report" "Deadline"
+        requires "Level" "ED" "TR" "Issue Tracking" "Date" "Implementation Report"
         group-types "wg"
     }
     status "PR" "W3C Proposed Recommendation" {


tidoust:
The contents in this folder is a copy of [`speced/bikeshed-boilerplate`](https://github.com/speced/bikeshed-boilerplate) (hence the *readonly* in the path) and only used in tests I believe. The PR rather needs to be made against the [`doctypes.kdl`](https://github.com/speced/bikeshed-boilerplate/blob/main/boilerplate/doctypes.kdl) file in `speced/bikeshed-boilerplate`, as that's the file that contains the live data that Bikeshed uses when it runs.

PR #3033, opened 2025-02-13
fix solidcg copyright includes

elf-pavlik:
We are switching to spec-prod and Nu HTML check didn't pass due to nested `<p>` tags.


tabatkins:
This edit needs to go into `bikeshed-boilerplate`, not here, but also, it shouldn't be erroring. Hm, I guess it's getting a `p` from the `[TITLE]` macro now? Let me poke at that.

tabatkins:
Ah, no, it's because the copyright include is put into a `<p>` in the W3C's `header.include`, got it.

I've gone ahead and applied the fix in the `bikeshed-boilerplate` repo for you.

PR #3104, opened 2025-05-19
Requalify "Informative References" into Non-Normative

### LINKED ISSUE #3077: "non-normative" vs "informative"
Occasionally, people get confused about the word "informative", as they read it in the sense of interesting and bringing new information, rather than the often intended "non-normative". Due to that, I've made a practice of saying "non-normative".

However, bikeshed generates some instances of "informative". One being the section title "Informative References". Could you change that to "Non-normative references"?

frivoal:
Informative is occasionally taken to be a subjective qualifier indicating that something is very interesting or bringing significant new information, instead of the intended contrast with normative.

Closes #3077


frivoal:
@tabatkins I wasn't sure: was I supposed to rebase the tests to go along with the proposed changes as I did, or is that supposed to be handled separately somehow?

PR #3135, opened 2025-07-16
add STATEMENT to index.console.txt

tantek:
TAG can and has published STATEMENTs


frivoal:
You should be able to make the failing test pass using `bikeshed test --rebase`, as documented in https://github.com/speced/bikeshed/blob/main/CONTRIBUTING.md#user-content-running-tests, though I don't know for sure whether the bikeshed maintainers want to do that themselves after accepting the PR or expect you to do it first.

tabatkins:
There shouldn't be any need for this, since the tests are generated with a snapshot of the boilerplates, not the live ones. I (very) occasionally update the test boilerplates, at which point *everything* will get rebased, but until then pure-boilerplate changes shouldn't cause any change here.

PR #3137, opened 2025-07-16
add STATEMENT, more specific Note track link to status.include

tantek:
add include-if paragraph for STATEMENT, use more specific Process fragment ID for Note track links


frivoal:
(also, I am unsure why the proposed fix is applied to the boilerplate of the mediawg only, rather than to all the wg's that have status text for notes, or for all W3C groups generally.

tantek:
Thanks @frivoal I have created a PR in the proper repo instead (https://github.com/speced/bikeshed-boilerplate/pull/165) and am closing this without change.

PR #3119, opened 2025-06-27
read and display data related to the CG spec transition plan if any

deniak:
The metadata file is expected to be located in the same directory as the spec itself and named `cg-monitor.json`.
Sample:
```json
{
  "transition_status": "Ready to transfer",
  "intended_implementations": [
    "Chrome",
    "Firefox",
    "Safari"
  ],
  "implementations": [
    "Chrome",
    "Firefox",
    "Safari"
  ],
  "intended_organization": "W3C",
  "intended_group": "webapps",
  "transition_plan": {
    "decision_to_migrate": null,
    "decision_to_receive": null,
    "ongoing_discussions": "https://github.com/w3c/webappswg/issues/118#issuecomment-2840981975"
  },
  "notes": [
    "The WebApps WG may add File and Directory Entries to the WD of File APIs after rechartering."
  ]
}
```

PR #3143, opened 2025-07-25
Add CG Transition Plan metadata, and build a note into Status based on that.

tabatkins:
Replaces #3119

### DIFF
--- a/bikeshed/boilerplate.py
+++ b/bikeshed/boilerplate.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import copy
 import dataclasses
+import json
 import os
 import re
 import subprocess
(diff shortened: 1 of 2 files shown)


tabatkins:
@deniak Just checking - is it intended to be the case that *every* cg-monitor.json file must contain a "transition_status", "intended_organization", and "intended_group" key, all plain text, and the "notes" key can potentially contain HTML? I've coded against those assumptions.

deniak:
> @deniak Just checking - is it intended to be the case that _every_ cg-monitor.json file must contain a "transition_status", "intended_organization", and "intended_group" key, all plain text, and the "notes" key can potentially contain HTML? I've coded against those assumptions.

Thanks for reworking the PR. Just to be clear, that feature is only at the stage of prototyping and we are still discussing the details with the WICG chairs so please, do not merge this yet.

Now to answer your questions, we do expect the json to contain the keys `transition_status` and `intended_organization`. `intended_group` might be optional depending on the targeted organization.
As for the `notes`, I had planed to allow the use of HTML to give editors the possibility to provide more details but that's certainly something we can revise if this is an issue.

Initially, we fought about including these properties as bikeshed metadata but we also want this to work with respec and we will also build other tools relying of these data, hence the json.

PR #3186, opened 2025-10-08
Make block-level markdown a Markup Shorthands options

foolip:
This will be enabled by HTML for a slight speed improvement, and not
having to carefully escape/indent certain things in the initial
conversion. We might later want to turn it on again.


tabatkins:
That's definitely a regression; all the specs using `markdown` aren't getting their inline markdown working. I expect you missed some spot where that markupShorthands value is checked, so it's still looking for `markdown` rather than `markdown-inline`.

Yup, `shorthands/oldShorthands.py` needs that switch, in two spots.

PR #3232, opened 2026-01-16
chore: Update Python base image to version 3.12 In Dockerfile

smessie:
Building the Dockerfile image failed due to stronger requirements in the dependency packages.
Using Python 3.12, building the image works again.


smessie:
Can we get this merged? @tidoust @tabatkins

PR #3294, opened 2026-04-18
ci: Group Dependabot updates monthly

nschonni:
Move the dependabot updates to a monthly cadency, and allow creating bundles updates for any minor/patch updates. Major updates will still get their own PRs


tabatkins:
You are an angel

PR #3303, opened 2026-05-29
update AOM copyright & spec status text

podborski:


tabatkins:
Oh, sorry, this needs to be updated in <https://github.com/speced/bikeshed-boilerplate/> instead

Issue #3319, opened 2026-08-31
Build MDN annotation panels on demand from a data attribute


zcorpan:
Bikeshed builds the whole MDN annotation panel body at generation time and hides it until the reader opens the `<details>`. For the HTML Standard that came to ~58k elements and ~1.8MB of markup, on a page whose load is ~72% style and layout, for content almost nobody expands.

wattsi now emits the body as one compact attribute per panel instead, hydrated on first open:

```html
<details class="mdn-anno before" data-mdn="v1|HTML/Element/video|9|3.5,3.1,3,10.5,79,12,9,?,3,?,?,?,?">
 <summary><b class=all-engines-flag title="This feature is in all current engines.">✔</b><span>MDN</span></summary>
</details>
```

That took `index-html` from 13.9MB to 12.2MB and from 264k to 206k elements. The builder is [`standard-mdn-annos.js`](https://resources.whatwg.org/standard-mdn-annos.js) and the format is documented at the top of `EncodeMDNAnnotations` in wattsi's `src/wattsi.pas`. Both were written to be generator-neutral, so `mdnspeclinks.py` could emit the same attribute and bundle the same script as `bikeshed/stylescript/mdn-anno.js`:

- Slots are positional and append-only, and an empty cell omits that row — which is what Bikeshed already does for a browser absent from `support`. The `nodejs` slot is there too.
- The `caniuse` field is optional, so Bikeshed can just not emit it.
- The version prefix means the format can change later without breaking archived output.

Two things it would cost Bikeshed, both worth knowing before deciding:

- The browser labels and engine-support strings are hard-coded in the script with no override hook, so they'd stop going through `_t()`. That was a deliberate simplification on the WHATWG side rather than an oversight, but it's a real regression for translated specs.
- `bikeshed/mdn/mdn-anno.css` has no `.unknown` or `.caniuse` rules. Bikeshed doesn't emit either today, so nothing breaks, but they'd be needed if it ever did.

The `<details>`/`<summary>` DOM and the `summary`/`[open]` CSS are Bikeshed's own, adopted unchanged — wattsi moved onto them as part of this, and `standard-shared-with-dev.css` now carries copies of those rules, verified to produce no computed-style change on a Bikeshed panel. Filed off the back of https://github.com/whatwg/wattsi/issues/169.

---

This is step 7 of https://github.com/whatwg/wattsi/issues/169

Generated by Claude.

PR #3324, opened 2026-09-11
Switch MDN annos to ship as just a code string, and get hydrated on load. #3319

tabatkins:
Regression: some features had a small description previously, but that seems to no longer be part of the code format.


bikeshed/mdn/mdn-anno.js
```diff
@@ -0,0 +1,202 @@
...
+// Builds MDN annotation panels on demand from the compact `data-mdn` attribute written by
+// the spec generator (wattsi's EncodeMDNAnnotations; Bikeshed's mdnspeclinks.py builds its
+// panels eagerly for now). Only the wire format is shared: the browser labels, the
+// engine-support strings and the render grouping all live here.
+//
+// Archived commit snapshots load this file forever, so every format version this script
+// has ever understood has to keep working. Evolve the format by adding a branch in
+// `hydrate`, never by redefining a version that has already shipped.
+
+const MDN_URL = 'https://developer.mozilla.org/en-US/docs/Web/';
```
zcorpan: This file needs to be updated to include https://github.com/whatwg/whatwg.org/pull/506/changes/305ad65b64a0b2c3c6bdf681b89d42f2030a880a , see comments in https://github.com/whatwg/wattsi/pull/170

tests/github/WICG/shape-detection-api/index-zh-cn.html
```diff
@@ -2100,6 +2074,213 @@ <h2 class="heading no-num no-ref settled" id="idl-index"><span class="content">I
...
+    version: caveat?.flag ? `${FLAG} ${version}` : version,
+    title: caveat?.title(slot),
+  };
+}
+
+// A browser row is the browser's name in one cell and its version in the next. The logo
+// comes from CSS, keyed off the browser id in the row's class.
+function browserRow(slot, cell) {
+  const { state, version, title } = readCell(cell, slot);
+  return element('span', { class: `${slot.id} ${state}` },
```
zcorpan: `bikeshed/mdn/mdn-anno.css` is missing a rule for `.unknown`.

bikeshed/mdn/mdnspeclinks.py
```diff
@@ -160,148 +154,150 @@ def panelsFromData(doc: t.SpecT, data: MdnDataT) -> list[t.ElementT]:
...
-                if "flags" in thisBrowserSupport:
-                    needsFlag = True
                 versionAdded = versionDetails["version_added"]
-                versionRemoved = None
+                if "prefix" in support or "alternative_name" in support:
+                    caveat = "$"
+                elif "partial_implementation" in support:
+                    caveat = "*"
+                elif "flags" in support:
+                    caveat = "^"
```
zcorpan: `support` should be `versionDetails` for these lines.

bikeshed/mdn/mdnspeclinks.py
```diff
@@ -160,148 +154,150 @@ def panelsFromData(doc: t.SpecT, data: MdnDataT) -> list[t.ElementT]:
...
-    h.appendChild(outer, h.E.span({}, browserFullName))
-    h.appendChild(outer, h.E.span(minVersionAttributes, flagSymbol + minVersion))
-    return outer
+    versionCode = "?"
+    if versionAdded is False:
+        versionCode = "-"
+    elif versionAdded is True:
+        versionCode = "!"
+    elif versionAdded:
+        versionCode = versionAdded
```
zcorpan: Regression: the edge-18 / ie-11 special cases were lost.

bikeshed/mdn/mdnspeclinks.py
```diff
@@ -160,148 +154,150 @@ def panelsFromData(doc: t.SpecT, data: MdnDataT) -> list[t.ElementT]:
...
 
-def addSupportRow(
+def encodeFeatureData(features: list[MdnFeatureT]) -> str:
+    fullCode = "v1|" + ("~".join(codeFromFeature(x) for x in features))
+    return fullCode
+
+
+def codeFromFeature(feature: MdnFeatureT) -> str:
+    #   feature  = slug "|" level "|" cells ["|" caniuse-feature "," caniuse-title]
+    code = cleanCodeText(feature["slug"]) + "|"
```
zcorpan: This needs a change similar to https://github.com/whatwg/wattsi/pull/170/changes/efcb2105d2922c7e2fdcac188a397282bd4a40d8 to fix the broken links issue.

bikeshed/mdn/mdnspeclinks.py
```diff
@@ -160,148 +154,150 @@ def panelsFromData(doc: t.SpecT, data: MdnDataT) -> list[t.ElementT]:
...
+    #   cell     = [caveat] body
+    #   caveat   = "*" partial | "^" needs a flag | "$" needs a prefix or alternative name | "@" mirrored
+    #   body     = "?" unknown | "-" unsupported | "!" supported, version unknown
+    #            | "=" version           supported since exactly this version, no trailing "+"
+    #            | version "-" version   supported over a range, rendered with an en dash
+    #            | version               supported since this version, rendered with a "+"
+
+    # Documentation for the "support" data:
+    # https://github.com/mdn/browser-compat-data/blob/main/schemas/compat-data-schema.md
+    support = feature["support"].get(browserCodeName)
```
zcorpan: This throws for `html-build --bikeshed` because some entries in `html.json` have `support: null`.

bikeshed/mdn/mdnspeclinks.py
```diff
@@ -160,148 +154,150 @@ def panelsFromData(doc: t.SpecT, data: MdnDataT) -> list[t.ElementT]:
...
-    return outer
+    versionCode = "?"
+    if versionAdded is False:
+        versionCode = "-"
+    elif versionAdded is True:
+        versionCode = "!"
+    elif versionAdded:
+        versionCode = versionAdded
+        if versionRemoved:
+            versionCode += "-" + versionRemoved
```
zcorpan: BCD allows `version_removed: true` which would cause this to throw.

bikeshed/mdn/mdn-anno.js
```diff
@@ -0,0 +1,202 @@
+'use strict';
+// File copied from Wattsi.
```
zcorpan: Better: whatwg/whatwg.org standard-mdn-annos.js

bikeshed/mdn/mdnspeclinks.py
```diff
@@ -160,148 +154,150 @@ def panelsFromData(doc: t.SpecT, data: MdnDataT) -> list[t.ElementT]:
...
+            return "0"
+        elif numEngines == 1:
+            return "1"
+        elif numEngines >= len(BROWSER_DATA["current"]):
+            return "9"
+    return ""
+
+
+def cellsFromFeature(feature: MdnFeatureT) -> str:
+    #   cells    = cell ("," cell)*  one per entry of BROWSER_DATA["cell_order], in that order. An
```
zcorpan:
```suggestion
    #   cells    = cell ("," cell)*  one per entry of BROWSER_DATA["cell_order"], in that order. An
```

bikeshed/mdn/mdnspeclinks.py
```diff
@@ -16,16 +16,62 @@ class MdnFeatureT(t.TypedDict, total=False):
...
+        partial_implementation: t.Literal[True]
+        flags: t.Any
+
     MdnDataT: t.TypeAlias = dict[str, list[MdnFeatureT]]
 
+BROWSER_DATA = {
+    "current": ["firefox", "safari", "chrome"],
+    "borrowed": ["opera", "edge_blink"],
+    "retired": ["edge", "ie"],
+    "mobile": [
```
zcorpan: Claude says these are not used.

bikeshed/mdn/mdnspeclinks.py
```diff
@@ -160,148 +154,150 @@ def panelsFromData(doc: t.SpecT, data: MdnDataT) -> list[t.ElementT]:
...
+    caveat = ""
+    if isinstance(support, dict):
+        if "version_added" in support:
+            versionAdded = support["version_added"]
+            if "prefix" in support or "alternative_name" in support:
+                caveat = "$"
+            elif "partial_implementation" in support:
+                caveat = "*"
+            elif "flags" in support:
+                caveat = "^"
```
zcorpan: I think flag > prefix > partial is better. Some features have 2 applied (partial + prefix, flag + prefix or flag + partial).

```suggestion
            if "flags" in support:
                caveat = "^"
            elif "prefix" in support or "alternative_name" in support:
                caveat = "$"
            elif "partial_implementation" in support:
                caveat = "*"
```

bikeshed/mdn/mdnspeclinks.py
```diff
@@ -160,148 +154,150 @@ def panelsFromData(doc: t.SpecT, data: MdnDataT) -> list[t.ElementT]:
     return panels
 
 
-def addSupportRow(
+def encodeFeatureData(features: list[MdnFeatureT]) -> str:
+    fullCode = "v1|" + ("~".join(codeFromFeature(x) for x in features))
+    return fullCode
```
zcorpan: Distinct BCD features can share an MDN page and identical support data. This could deduplicate identical records. (whatwg/wattsi#170 now does.)

```suggestion
def encodeFeatureData(features: list[MdnFeatureT]) -> str:
    # Deduplicate
    codes = dict.fromkeys(codeFromFeature(x) for x in features)
    return "v1|" + "~".join(codes)
```

bikeshed/mdn/mdnspeclinks.py
zcorpan: Also the same order as 266–271:

```suggestion
                if "flags" in versionDetails:
                    caveat = "^"
                elif "prefix" in versionDetails or "alternative_name" in versionDetails:
                    caveat = "$"
                elif "partial_implementation" in versionDetails:
                    caveat = "*"
```

bikeshed/mdn/mdnspeclinks.py
```diff
@@ -16,16 +16,62 @@ class MdnFeatureT(t.TypedDict, total=False):
...
+        "ie",  # retired
+        "firefox_android",
+        "safari_ios",
+        "chrome_android",
+        "webview_android",
+        "samsunginternet_android",
+        "opera_android",  # mobile
+        "nodejs",  # JS
+    ],
+}
```
zcorpan: Suggestion to resolve "mirror" like in whatwg/wattsi#171
```suggestion
}

# Which browser BCD copies a "mirror"ed browser's support data from. BCD's "edge" isn't
# here: Chromium Edge is reported separately as "edge_blink", so see cellFromFeature().
MIRROR_SOURCES = {
    "chrome_android": "chrome",
    "webview_android": "chrome",
    "samsunginternet_android": "chrome_android",
    "opera": "chrome",
    "opera_android": "chrome_android",
    "firefox_android": "firefox",
    "safari_ios": "safari",
}
```

bikeshed/mdn/mdnspeclinks.py
```diff
@@ -160,148 +154,150 @@ def panelsFromData(doc: t.SpecT, data: MdnDataT) -> list[t.ElementT]:
...
-                minVersion = "None"
-    browserFullName = nameFromCodeName[browserCodeName]
-    h.appendChild(
-        supportData,
-        browserCompatSpan(browserCodeName, browserFullName, statusCode, minVersion, needsFlag),
-    )
+    elif support == "mirror":
+        # A derived browser whose support matches the "upstream" one.
+        # Not handled here, and newer BCD data normalizes it away.
+        caveat = "@"
```
zcorpan: Suggestion to resolve "mirror" like in whatwg/wattsi#171 (also see comment for line 60)

```suggestion
    elif support == "mirror":
        # Support is whatever the browser this one derives from has. BCD maps the version
        # through release tables this data doesn't carry, so only report the state. A
        # mirrored legacy "edge" means the feature came after it, as Chromium Edge's
        # support is reported separately as "edge_blink".
        if browserCodeName == "edge":
            return "-"
        source = MIRROR_SOURCES.get(browserCodeName)
        upstream = cellFromFeature(feature, source).lstrip("$*^@") if source else ""
        if upstream in ("", "?"):
            return "@?"
        return "@-" if "-" in upstream else "@!"
```

zcorpan:
Comment by Claude:

> whatwg/wattsi#170 and whatwg/whatwg.org#506 now add an optional fifth field to `v1`, for when several records in a panel share an MDN article but not their support data, which a reader otherwise can't tell apart (whatwg/html#12987):
>
> ```
> feature = article "|" level "|" cells ["|" caniuse ["|" label]]
> ```
>
> A labelled record without caniuse data gets an empty caniuse field, so the script now checks `if (caniuse)` rather than `caniuse !== undefined`, and shows the label after the article link, e.g. `Attributes/disabled (<textarea>)`. Syncing `mdn-anno.js` with whatwg/whatwg.org#506 picks both up, and a script without them just ignores the field.
>
> On Bikeshed's data this would label 316 records in 51 panels, mostly WebGL methods, where each article has a `WebGLRenderingContext` and a `WebGL2RenderingContext` record with different support. Below is a version of `encodeFeatureData` (lines 157–159) that does it, and also covers the dedup suggested above. Only records that have a `filename` get labelled. The label is the element for `html/elements/*` or the file's name otherwise, then the file path if that doesn't tell them apart (`at-rules/namespace` vs `selectors/namespace`), then BCD's name, and nothing if none of those do. I checked it against all 294 data files: no group gets a repeated label, and stripping the labels gives exactly the deduped output.
>
> <details><summary>Python</summary>
>
> ```python
> def encodeFeatureData(features: list[MdnFeatureT]) -> str:
>     # Distinct BCD features can share an MDN page. Merge the ones that encode identically,
>     # and where one article still has several records, label each with what it's for.
>     records: dict[str, list[MdnFeatureT]] = {}
>     for feature in features:
>         records.setdefault(codeFromFeature(feature), []).append(feature)
>     # Entries without a "filename" aren't from BCD itself, so there's nothing to label them with.
>     fromBcd = {code: [f for f in fs if f.get("filename")] for code, fs in records.items()}
>     groups: dict[str, list[str]] = {}
>     for code, bcd in fromBcd.items():
>         if bcd:
>             groups.setdefault(code.split("|", 1)[0], []).append(code)
>     labels: dict[str, str] = {}
>     for codes in groups.values():
>         if len(codes) > 1:
>             labels |= distinctLabels(codes, fromBcd)
>     return "v1|" + "~".join(
>         code + ("|" if code.count("|") == 2 else "") + "|" + cleanCodeText(labels[code]) if code in labels else code
>         for code in records
>     )
>
>
> def distinctLabels(codes: list[str], fromBcd: dict[str, list[MdnFeatureT]]) -> dict[str, str]:
>     # Name the features the first way that tells every record apart: by element or file,
>     # then by file path, then by BCD feature name. If none does, a label wouldn't help.
>     for describe in (featureLabel, featurePath, lambda f: f.get("name", "")):
>         labels = {code: ", ".join(dict.fromkeys(describe(f) for f in fromBcd[code])) for code in codes}
>         if len(set(labels.values())) == len(labels):
>             return labels
>     return {}
>
>
> def featureLabel(feature: MdnFeatureT) -> str:
>     # The element for html/elements/*, otherwise the file's own name, e.g. the interface for api/*.
>     filename = feature.get("filename", "")
>     name = filename.rsplit("/", 1)[-1].removesuffix(".json")
>     return f"<{name}>" if filename.startswith("html/elements/") else name
>
>
> def featurePath(feature: MdnFeatureT) -> str:
>     # e.g. "at-rules/namespace" rather than "selectors/namespace".
>     return feature.get("filename", "").split("/", 1)[-1].removesuffix(".json")
> ```
>
> </details>
>
> Wattsi also warns about entries without a `filename` that shadow a BCD one, as they're probably stale in mdn-spec-links' `.local/html.json` (w3c/mdn-spec-links#854). Maybe worth doing here too.

bikeshed/mdn/mdnspeclinks.py
tabatkins: Ughhhhhhhhh I hate how underdocumented the exact data shapes of that data is.

bikeshed/mdn/mdnspeclinks.py
tabatkins: More hate for underdocumented data shapes.

bikeshed/mdn/mdnspeclinks.py
tabatkins: Eh didn't think it was that important, but sure, added back.

bikeshed/mdn/mdnspeclinks.py
tabatkins: I'm fine with that order. Wrt your last sentence, do you mean that a cell can have multiple caveats? Or just that a support entry can have multiple caveat-causing things, and you prefer this particular order of processing?

bikeshed/mdn/mdnspeclinks.py
tabatkins: Just checking with your own handling - this suggested mirror handling seems wrong at the end. The `if "-" in upstream` captures both a version of `"-"` (unsupported) and a version range (supported, but with unknown versions in this browser). I'm correctly checking for an explicit "-" now, but make sure your own code is handling this correctly.

